PR #18, opened 2019-10-01
Improving mesh separating

RomanFrom710:
Handle `targets` primitive property when splitting a mesh because it may contain buffer accessors indexes as well.

Also fixed `if (primitive.indices)` condition (because `indices` can be `0`) and the test case that was affected by this bug.


donmccurdy:
Thanks!

PR #19, opened 2019-10-01
Fixed buffer view removing

RomanFrom710:
`bufferView.byteOffset > bufferView.byteOffset` is always false, so byte offset of the rest of buffers isn't aligned properly.

Also we need to align only buffer views that point to the same buffer as the buffer view being removed, so added check for it as well.


donmccurdy:
Ouch, good find. 😅

Thanks!

PR #71, opened 2020-08-15
Migrate tests to TypeScript

donmccurdy:
Remaining:
- [ ] Ensure that browser tests work (with tsify?)
- [x] Ensure that coverage works (see https://github.com/benwhite-deltas/typescript-tape-nyc-example)
- [x] Migrate all the tests.


donmccurdy:
Tests and coverage work. But so far I can't get the tests to compile and run in the browser, at least with browserify and tsify.

PR #103, opened 2020-09-24
Stop direct access to json tags

user-ops:
images/buffers may not be specified in json according to spec, we can add an empty array to allow the spread operation to occur without throwing


donmccurdy:
Thanks!

PR #172, opened 2021-01-26
Add variant transform.

### LINKED ISSUE #64: Add 'variants' lib and CLI function
See https://github.com/KhronosGroup/glTF/pull/1681, similar to https://github.com/facebookincubator/glTFVariantMeld. Would be added to the `lib` package and available through the CLI. Not sure whether the existing meld API is the right one, maybe this should take a config file pointing to external textures, and a single glTF, instead?

donmccurdy:
Fixes #64.

As it is, this extension does not attempt to detect reuse of materials or their
extensions. Switching variants will therefore replace _all_ materials, and that's not
exactly ideal. That is harder to fix, though, and requires detecting duplicate materials.
This could be accomplished by extending `dedup`, or implementing a general `equals`
or `hashKey` function. Extensions, in particular, complicate that.

Remaining:

- [ ] Improve variant naming logic
- [ ] Documentation
- [ ] Unit tests

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -6,7 +6,7 @@ import { gzip } from 'node-gzip';
 import { program } from '@caporal/core';
 import { Logger, NodeIO, PropertyType } from '@gltf-transform/core';
 import { ALL_EXTENSIONS } from '@gltf-transform/extensions';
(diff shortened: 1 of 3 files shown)


donmccurdy:
This has been open for about a year, and I haven't had time to complete it. I don't know when or if that will change, so I'm closing the PR for now.

>  That is harder to fix, though, and requires detecting duplicate materials.

This part is much easier now, since there is a `material.equals(otherMaterial)` method that can detect these things.

Issue #191, opened 2021-02-19
Including identity transforms in json output


dbuck:
Question: I was working with a few files, and I added some new nodes. After writing out the gltf files, I was not expecting to see new identity values for the node position/rotation/etc.

Initial output from the original obj2gltf, which was the source before using in gltf-transform:
```
  "nodes": [
    {
      "name": "my_awesome_node_name",
      "mesh": 0
    },
```

Output gltf after appending some extra info:
```
  "nodes": [
    {
      "name": "my_awesome_node_name",
      "extras": {
        "room": 7,
        "classification": "unknown",
      },
      "translation": [ 0, 0, 0 ],
      "rotation": [ 0, 0, 0, 1 ],
      "scale": [ 1, 1, 1 ],
      "mesh": 0
    },
```

It doesn't seem like properties/node.getTranslation() should change, but inserting that behavior into the writer also seems like an awkward place?

PR #192, opened 2021-02-19
[core] - test for node output including optional identity values

dbuck:
Adds a failing test to go along with question from #191

I can fill in a 'fix' if you have a preference about where to address the issue (Still getting bearings in the project while testing it out)


donmccurdy:
Thank you! I appreciate the test, and agree that it would be good to avoid writing the "default" values of glTF properties. If you'd be willing to try to fix the issue, I think a change in `writer.ts` doing something like this would be perfect:

```ts
if (!MathUtils.eq(node.getTranslation(), [0, 0, 0])) {
  nodeDef.translation = node.getTranslation();
}
```

donmccurdy (approved):
Looks great, thank you!

PR #202, opened 2021-03-19
[core] Fix missing default scene from document root.

ptc-mrucci:
The `scene` field at the root of the document is [part of the glTF specs](https://github.com/KhronosGroup/glTF/blob/2354846cbab8f8081ad80dcb9d5371f767ea0dbf/specification/2.0/schema/glTF.schema.json#L118).
This PR fixes the reader/writer to include the `scene` field and the internal data structures so that the default scene is preserved when transforming or copying.


packages/core/src/properties/root.ts
```diff
@@ -68,6 +68,8 @@ export class Root extends Property {
 	@GraphChildList private skins: Link<Root, Skin>[] = [];
 	@GraphChildList private textures: Link<Root, Texture>[] = [];
 
+	private scene?: number;
```
donmccurdy, line 71: Let's make this a `@GraphChild` property, similar to this example:

https://github.com/donmccurdy/glTF-Transform/blob/7be7f9ddaab098af79b1eea3b67adac3d9c453ed/packages/core/src/properties/skin.ts#L22

The advantage of this is that if the scene is disposed, or the scene order changes, the references are maintained. For example:

```ts
root.setDefaultScene(scene);
root.getDefaultScene(); // -> scene

scene.dispose();

root.getDefaultScene(); // -> null
```

packages/core/src/properties/root.ts
```diff
@@ -180,6 +182,21 @@ export class Root extends Property {
 		return this.scenes.map((p) => p.getChild());
 	}
 
+
+	/**
+	 * Set the default scene ID associated with this root.
+	 * @hidden
+	 */
+	public _setDefaultScene(sceneId?: number): this {
```
donmccurdy, line 190: I think this method can be public, and just remove the method from the Document. Similar to `getExtras()` and `setExtras(...)` on the Root.

packages/core/src/io/reader-context.ts
```diff
@@ -20,6 +20,7 @@ export class ReaderContext {
 	public skins: Skin[] = [];
 	public animations: Animation[] = [];
 	public scenes: Scene[] = [];
+	public scene?: number;
```
donmccurdy, line 23: It doesn't seem like `context.scene` is actually used, maybe this could be omitted until it's needed? If your project needs it for some reason that's OK too.

packages/core/src/document.ts
```diff
@@ -153,6 +153,9 @@ export class Document {
 			thisProp.copy(otherProp, resolve);
 		}
 
+		// 5. Copy default scene
+		this.setDefaultScene(other.getRoot().getDefaultScene());
```
donmccurdy, line 157: With the suggestion below of making `setDefaultScene` public on the Root, this logic could be moved into `root.copy()`.

donmccurdy (commented):
Hi, thanks for this PR and for including tests! I agree that maintaining the default scene is worthwhile and should be supported. I have a few suggestions below about how to expose the methods, but I think we can get this merged pretty quickly. :)

ptc-mrucci:
Thanks for the quick feedback, I've addressed the first batch of requests and continue working on the rest.

packages/core/src/properties/root.ts
ptc-mrucci: Ah, make sense, I didn't fully understand the whole GraphChild and Link logic and was lazily trying to skip that part in the hope it wouldn't be necessary :D.

Just to clarify, you would like to define a `scene` attribute that is a link/reference to a Scene object in the `scenes` list, so that the actual value is determined at runtime when writing the json for example?
In that case, would I need a `sceneIndexMap`  similar to the `nodeIndexMap` in the writer context?
Also, once the `root.scene` is not a number anymore, would the setDefaultScene of the document and root accept a Scene reference or a number (and then map the number to a Scene)?
For consistency with other methods seems like I should use the former.

donmccurdy:
Thank you!

PR #209, opened 2021-04-02
[lib] Improve attribute quantization. (fixes #208)

### LINKED ISSUE #208: Improve quantization
Current implementation discontinuous at 0, for bits >8 and <16. The most obvious 
way of fixing that would be...

Math.round(el[j] * (quantScale - 1)) * storeScale;

...but then we don't maintain 1 -> (2^storeScale - 1) at the extremes. Could
also consider doing a bit more rounding...

Math.round(el[j] * (quantScale - 1)) * (storeScale + K)

...but I'm not at all sure that's right, and if our goal was to have some bit-
level consistency in the result we've surely erased it. Moreover, whatever we
implement needs to "just work" for bits=8 or bits=16.

donmccurdy:
/cc @lexaknyazev

### DIFF
--- a/packages/lib/src/quantize.ts
+++ b/packages/lib/src/quantize.ts
@@ -222,6 +222,8 @@ function quantizeAttribute(
 	const storageBits = ctor.BYTES_PER_ELEMENT * 8 - signBits;
 
 	const scale = Math.pow(2, quantBits) - 1;
+	const lo = storageBits - quantBits;
+	const hi = 2 * quantBits - storageBits;
 
 	for (let i = 0, di = 0, el = []; i < attribute.getCount(); i++) {
 		attribute.getElement(i, el);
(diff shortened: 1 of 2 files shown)


packages/lib/src/quantize.ts
```diff
@@ -210,17 +217,23 @@ function quantizeAttribute(
...
 
 	for (let i = 0, di = 0, el = []; i < attribute.getCount(); i++) {
 		attribute.getElement(i, el);
 		for (let j = 0; j < el.length; j++) {
-			dstArray[di++] = Math.round(el[j] * quantScale) * storeScale - 1 * Math.sign(el[j]);
+			// Map [0.0 ... 1.0] to [0 ... scale].
+			let value = Math.round(Math.abs(el[j]) * scale);
+
+			// Replicate msb to missing lsb.
+			value = (value << (storageBits - quantBits)) | (value >> (2 * quantBits - storageBits));
```
lexaknyazev, line 233: Consider moving `storageBits - quantBits` and `2 * quantBits - storageBits` out of loops for better readability (and potentially faster execution).

PR #234, opened 2021-04-20
Upgrade to typedoc v0.20

### LINKED ISSUE #182: Update to Typedoc v0.20
v0.20 is a breaking change, and removes several CLI options I'm using now. It also has a different output structure and may break compatibility with my current (customized) theme. Release notes:

https://github.com/TypeStrong/typedoc/releases/tag/v0.20.0

Not sure how much work will be required to get this updated. Might be straightforward. If not, I'm tempted to stop using its theming system entirely and just push the JSON output into custom templates myself. It would be nice to be able to automatically generate documentation for the `lib` and `extensions` packages, whereas only `core` has auto-generated docs now.

donmccurdy:
Fixes #182.

Several good things here: Typedoc will now let us document the `extensions` and `lib` packages, not just `core`, on the same site. And the CLI seems to be a bit faster and nicer. But, there are some breaking things to work through here:

- [ ] Various things missing from the theme. Review https://github.com/mipatterson/typedoc-plugin-pages/issues/47 and https://github.com/mipatterson/typedoc-plugin-pages/pull/41.
- [ ] The markdown pages I've tried to include (Concepts, Library, Extensions, ...) are getting dropped. Might need to lean on https://github.com/mipatterson/typedoc-plugin-pages here.

Maybe also have a look at the `@category` tag usage, it might get rid of the need to manually write the TOC here. https://github.com/TypeStrong/typedoc/issues/564.


donmccurdy:
A few more thoughts:

- Should probably start by just upgrading, and add the other packages to the docs in a separate PR.
- When that happens, links to many existing pages will change, so we may want Vercel to handle the redirects.

donmccurdy:
Waiting for https://github.com/TypeStrong/typedoc/issues/1580.

PR #246, opened 2021-05-11
Support for latest KTX Software build, fix build dependencies

dxdc:
- Fix build dependencies
- Prevent error with latest KTX Software build `error: Requires KTX-Software >= 4.0.0, found v4.0.0-5-gca6f6e7d.`

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -29,7 +29,7 @@
     "csv-stringify": "^5.6.2",
     "draco3dgltf": "^1.4.1",
     "gl": "^4.9.0",
-    "gltf-validator": "^2.0.0-dev.3.2",
+    "gltf-validator": "^2.0.0-dev.3.3",
     "ktx-parse": "^0.2.1",
     "markdown-table": "^2.0.0",
     "mikktspace": "^0.2.0",
@@ -47,5 +47,8 @@
     "package.json",
     "package-lock.json"
   ],
-  "gitHead": "9f80b1a7b25076e9b6d1dd5c4702daa3ce1b67ae"
+  "gitHead": "9f80b1a7b25076e9b6d1dd5c4702daa3ce1b67ae",
+  "devDependencies": {
+    "microbundle": "^0.13.0"
+  }
 }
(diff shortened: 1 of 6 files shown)


donmccurdy:
Hi, and thanks for letting me know about this. It looks like the current version policy of KTX-Software is non-standard, I'm looking into whether that could be changed (see https://github.com/donmccurdy/glTF-Transform/issues/2170). Let's hold off on the change here for now, since (with the current version policy) this would allow older `alpha` releases that aren't actually supported.

Aside — I don't believe the build changes should have been necessary, and would like to improve the [contributing guide](https://gltf-transform.donmccurdy.com/contributing.html) if it didn't work for you. My recommended process would be, from the root the project (not `packages/cli`) —

```shell
yarn install
yarn dist

cd packages/cli
yarn link

# test local CLI
gltf-transform -h
```

A similar process might work with the `npm` client rather than `yarn`, but it didn't support workspaces the last time I tried this and so I'm not sure.

donmccurdy:
> I was able to build it properly now with your suggestions

Awesome, glad to hear.

> Fair enough, but, is there a non-working 4.0.0.X version of KTX-Software?

Yeah, the 4.0.0-beta versions won't work, and are pretty recent so I suspect some users still have them. Unintuitively 4.0.0-beta is technically "newer" than 4.0.0-5 according to semver version policy, even though it is actually older and less stable, so I'm hoping KTX-Software will stable versions as 4.0.x instead.

donmccurdy:
Hm, with the latest KTX-Software release I'm not able to reproduce this, and the version string uses a tilde instead of a dash, which is good: `v4.0.0~6`. Not entirely sure if that's a difference between the releases, or between platforms, or what. For now I'm just going to loosen the version check to a warning and not an error, so it doesn't stop you from going ahead with whatever you need to do: https://github.com/donmccurdy/glTF-Transform/commit/904607db0939583726658b6491d8de3a9713d31b

donmccurdy:
Thanks for looking into the issue and please let me know if you have any more trouble here!

PR #281, opened 2021-06-30
Remove ao() command and dependencies.

### LINKED ISSUE #164: Consider removing 'ao' from the CLI
Dependency on headless-gl makes installation unnecessarily complicated for any platform where prebuilt binaries aren't available. Might need to remove the `ao` command from the CLI until I have a better alternative for supporting that, and just let users handle it with scripts.

donmccurdy:
- Affected deps: gl, regl, geo-ambient-occlusion.

The function (and, mostly, its dependencies) represent a pretty large chunk of bundle size and installation time, particularly on platforms where precompiled 'gl' builds don't exist. Doesn't seem worth it / I'm not aware that many (any?) users are coming to this library for AO baking.

Fixes #164.


mrpeu:
Is there another way to bake AO with ThreeJS?

donmccurdy:
three-gpu-pathtracer can bake AO better than my implementation in glTF Transform could. It isn't fully wired up for embedding the AO map into a glTF file, but it shouldn't be a huge lift to set that up if it's what you're looking for. See:

- https://github.com/gkjohnson/three-gpu-pathtracer
- https://github.com/gkjohnson/three-gpu-pathtracer/pull/32

Or Blender is always a good option. :)

mrpeu:
Awesome. Thank you for the follow up, and for the original work of course

Issue #106, opened 2020-10-08
EXT_meshopt_compression


Inviz:
I want to compress my files with `gltfpack` but unfortunately, because it uses its custom decoder, i can not feed compressed files to gltf-transform. The issue I have with `gltfpack` is that it's a bit too rigid and sometimes does unwanted things like hyperoptimizing animation tracks. I would love to be able to merge compressed and un-compressed files together with gltf-transform.

Inviz:
I guess for my needs even gluing two glb files together (of them one compressed with EXT_meshopt_compression) would be good enough. 

Another issue with `gltfpack` is that it does not support multiple scenes, so it lumps them all into one. So i have to do turn scenes into nodes with gltf-transform before feeding to gltfpack, which is pretty ugly and unfortunate. But i made peace with it

donmccurdy:
> The issue I have with gltf-transform is that it's a bit too rigid and sometimes does unwanted things like hyperoptimizing animation tracks

You mean gltfpack here, correct?

I need to think about how to support this... one option would be to decompress the file while loading and then recompress it while writing, which would probably depend on https://github.com/zeux/meshoptimizer/issues/164, or something like it. Currently `KHR_draco_mesh_compression` is decompressed while loading, but I haven't implemented the compression side for export yet. But repeated lossy compression doesn't sound ideal, with either compression method.

I'd also like to support quantization — one of the compression techniques that gltfpack uses — directly in this tool see #59. 

Probably not a quick fix here, sorry. You might get a more helpful response requesting that gltfpack provide some options for individual optimizations in isolation? Otherwise I may need to wait and see what happens with https://github.com/zeux/meshoptimizer/issues/164.

Inviz:
> You mean gltfpack here, correct?
Yes, I apologize. I mean gltfpack

> But repeated lossy compression doesn't sound ideal, with either compression method.
Yeah for sure. If re-packing means re-quanitifying the files, then it will lead to even more losses. Not good, for sure. I guess just gluing files together would work. However un-packing and re-packing the EXT_meshopt_compression does not sounds like a lossy operation. So i think it could be possibly to decode those. And then `gltfpack` has `-noq`uantify option, so that it would not want to re-quanitify file that was quantified previously and glued together with `gltf-transform`, it could work well. So on your end there would only need to be a decoding.

At least that's what I understand! 


> I'd also like to support quantization — one of the compression techniques that gltfpack uses — directly in this tool see #59.
[...]

Issue #178, opened 2021-02-05
First argument to DataView constructor must be an ArrayBuffer when using inspect()


duncan-vrify:
I'm trying to implement `gltf-transform` into my Vue project in order to make use of the `inspect()` method for my gltf/glb files. 

This is what I'm trying to do 
```
import { WebIO } from '@gltf-transform/core'
import { inspect } from '@gltf-transform/lib'

// data.gltf === a URL at which the resource can be achieved (I can link it if necessary)
const gltfFile = data.gltf
const io = new WebIO()
const doc = await io.read(gltfFile)
inspect(doc)
```

But I seem to be getting this error

```
caught TypeError: First argument to DataView constructor must be an ArrayBuffer
    at new DataView (<anonymous>)
    at Function.m (core.modern.js:1)
    at Function.getSize (core.modern.js:1)
    at lib.modern.js:1
    at Array.map (<anonymous>)
    at q (lib.modern.js:1)
    at k (lib.modern.js:1)
    at _callee$ (_modelId.vue:180)
    at tryCatch (runtime.js:63)
    at Generator.invoke [as _invoke] (runtime.js:293)
    at Generator.next (runtime.js:118)
    at asyncGeneratorStep (asyncToGenerator.js:3)
    at _next (asyncToGenerator.js:25)
```

Any ideas as to what I'm doing wrong?

(Apologies if this isn't the right place to post this, let me know if it would be better suited for somewhere else)

donmccurdy:
> `const gltfFile = data.gltf`

So `gltfFile` is a string, correct? Does it have a `.gltf` or `.glb` extension? I don't think I will need access to the actual file here. 

It might be worth looking at your browser's network tab in the developer tools, to see what URL is being requested and whether it's finding a file (HTTP 200) or not (HTTP 404). Some web frameworks have development servers that handle static files differently.

duncan-vrify:
> 
> 
> > `const gltfFile = data.gltf`
> 
> So `gltfFile` is a string, correct? Does it have a `.gltf` or `.glb` extension? I don't think I will need access to the actual file here.
> 
> It might be worth looking at your browser's network tab in the developer tools, to see what URL is being requested and whether it's finding a file (HTTP 200) or not (HTTP 404). Some web frameworks have development servers that handle static files differently, e.g. Parcel usually needs the `parcel-plugin-static-files-copy` plugin to host any sort of binary file.

Yes, it's a string with a `.glb` extension

[...]

PR #313, opened 2021-07-23
Improve extension API to support buffer view providers.

donmccurdy:
Prerequisite for `EXT_meshopt_compression` (#106).

### DIFF
--- a/packages/core/src/io/platform-io.ts
+++ b/packages/core/src/io/platform-io.ts
@@ -74,7 +74,8 @@ export abstract class PlatformIO {
 			if (!resource.uri) {
 				const isGLBBuffer = isGLB && index === images.length;
 				if ((resource as GLTF.IImage)['bufferView'] === undefined && !isGLBBuffer) {
-					throw new Error('Missing resource URI.');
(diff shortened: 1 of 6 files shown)


packages/core/src/io/platform-io.ts
```diff
@@ -74,7 +74,8 @@ export abstract class PlatformIO {
 			if (!resource.uri) {
 				const isGLBBuffer = isGLB && index === images.length;
 				if ((resource as GLTF.IImage)['bufferView'] === undefined && !isGLBBuffer) {
-					throw new Error('Missing resource URI.');
+					// TODO(DO NOT SUBMIT): There was a reason this threw an error...
+					console.warn('PlatformIO: Missing resource URI.');
```
donmccurdy, line 78: Reasoning for this error was in https://github.com/donmccurdy/glTF-Transform/issues/178 and https://github.com/donmccurdy/glTF-Transform/pull/179, I guess I'll need to find a clearer way to detect missing resources without knowledge in this module of `EXT_meshopt_compression`.

PR #321, opened 2021-08-06
Add reorder function.

donmccurdy:
Implements a `reorder()` function, using Meshoptimizer for improving vertex locality and compression. Pre-processing step for Meshopt compression (#106).

Example:

```js
import { MeshoptEncoder } from './path/to/meshopt_encoder.module.js';
import { reorder, quantize } from '@gltf-transform/functions';

await document.transform(
  reorder({target: 'size', encoder: MeshoptEncoder}),
  quantize(),
  ...
);
```

Remaining:

- [x] Documentation
- [x] Clean up TS typings
- [x] Unit tests

### DIFF
--- a/packages/functions/src/reorder.ts
+++ b/packages/functions/src/reorder.ts
@@ -1,0 +1,153 @@
+import { Accessor, Document, GLTF, Primitive, PropertyType, Transform } from '@gltf-transform/core';
+import { prune } from './prune';
+import { SetMap } from './utils';
+import type { MeshoptEncoder } from 'meshoptimizer';
+
+const NAME = 'reorder';
+
+/** Options for the {@link reorder} function. */
+export interface ReorderOptions {
+	/** MeshoptEncoder instance. */
+	encoder?: typeof MeshoptEncoder,
+	/**
+	 * Whether the order should be optimal for transmission size (recommended for Web)
+	 * or for GPU rendering performance. Default is 'size'.
+	 */
+	target?: 'size' | 'performance',
+}
+
+const REORDER_DEFAULTS: Required<Omit<ReorderOptions, 'encoder'>> = {
+	target: 'size',
+};
+
+interface LayoutPlan {
+	indicesToMode: Map<Accessor, GLTF.MeshPrimitiveMode>;
+	indicesToAttributes: SetMap<Accessor, Accessor>;
+	attributesToPrimitives: SetMap<Accessor, Primitive>;
+}
+
+/**
+ * Optimizes {@link Mesh} {@link Primitive Primitives} for locality of reference. Choose whether
+ * the order should be optimal for transmission size (recommended for Web) or for GPU rendering
+ * performance. Requires a MeshoptEncoder instance from the Meshoptimizer library.
+ *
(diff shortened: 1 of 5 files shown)


donmccurdy:
@zeux I'm curious if you have plans of publishing the encoder and/or decoder to NPM? And if not, any objection to my putting the type definitions into the third-party [`@types/*`](https://github.com/DefinitelyTyped/DefinitelyTyped/) repository? gltf-transform is a monorepo containing several NPM packages, and unfortunately I can't share type definitions between those packages (`@gltf-transform/cli` and `@gltf-transform/functions` in this case) without (a) duplicating them or (b) importing them as a package. TypeScript users of this library may run into similar issues.

Any suggestions on this PR (e.g. improvements to the documentation) are also welcome. Output of `--help` on the CLI looks something like this:

```
$ gltf-transform reorder -h

gltf-transform 0.12.7 — Commandline interface for the glTF-Transform SDK.

  USAGE — reorder

    ▸ gltf-transform reorder <input> <output> [OPTIONS...]


  Optimize vertex data for locality of reference.

  Choose whether the order should be optimal for transmission size (recommended for Web) or for GPU
  rendering performance. When optimizing for transmission size, reordering is expected to be a pre-
  processing step before applying Meshopt compression and lossless supercompression (such as gzip or
  brotli). Reordering will only reduce size when used in combination with other compression methods.

  Based on the meshoptimizer library (https://github.com/zeux/meshoptimizer).


  ARGUMENTS

    <input>                              Path to read glTF 2.0 (.glb, .gltf) model
    <output>                             Path to write output

  OPTIONS

    --target                             Whether to optimize for transmission size or GPU
                                         performance
                                         one of "size","performance", default: "size"

  GLOBAL OPTIONS

    -h, --help                           Display global help or command-related help.
    -V, --version                        Display version.
    -v, --verbose                        Verbose mode: will also output debug messages.
    --vertex-layout <layout>             Vertex layout method
                                         one of "interleaved","separate", default:
                                         "interleaved"
```

packages/functions/src/reorder.ts
```diff
@@ -0,0 +1,152 @@
...
+
+const NAME = 'reorder';
+
+/** Options for the {@link reorder} function. */
+export interface ReorderOptions {
+	/** MeshoptEncoder instance. */
+	encoder?: MeshoptEncoder,
+	/**
+	 * Whether the order should be optimal for transmission size (recommended for Web)
+	 * or for GPU rendering performance. Default is 'compression'.
```
donmccurdy, line 13: TODO: 'compression' -> 'size'

zeux:
Yeah I was planning to publish both encoder and decoder as `meshoptimizer` NPM package (on the account of this being easier for me to update, and presumably any size-conscious web app would use webpack or smth similar that would strip unneeded files). Would that work?

donmccurdy:
That'd be perfect — publishing a single package should be fine here as long as an ES Module entrypoint is available.

zeux:
Published as https://www.npmjs.com/package/meshoptimizer, please let me know if anything is amiss.

PR #323, opened 2021-08-11
Support EXT_meshopt_compression encoding

donmccurdy:
Remaining: 

- [x] ~~Resolve various issues (normals, morph targets) on `-c` output~~
- [x] ~~Refactor into 'encoder.ts' and 'decoder.ts' files~~
- [x] ~~Unit tests~~
- [x] ~~Documentation~~
- [x] ~~Clean up duplicated buffer view target code~~
- [x] ~~Misc TODO items~~
- [ ] Fully implement `-cc` (filters)

### DIFF
--- a/packages/core/src/io/writer-context.ts
+++ b/packages/core/src/io/writer-context.ts
@@ -1,12 +1,23 @@
 import { Format } from '../constants';
 import { Document } from '../document';
 import { JSONDocument } from '../json-document';
-import { Accessor, Buffer, Camera, Material, Mesh, Node, Property, Skin, Texture, TextureInfo } from '../properties';
+import { Accessor, AttributeLink, Buffer, Camera, IndexLink, Material, Mesh, Node, Property, Skin, Texture, TextureInfo } from '../properties';
 import { GLTF } from '../types/gltf';
 import { ImageUtils, Logger } from '../utils';
 import { WriterOptions } from './writer';
 
 type PropertyDef = GLTF.IScene | GLTF.INode | GLTF.IMaterial | GLTF.ISkin | GLTF.ITexture;
+
+enum BufferViewTarget {
+	ARRAY_BUFFER = 34962,
+	ELEMENT_ARRAY_BUFFER = 34963,
+}
+enum BufferViewUsage {
+	ARRAY_BUFFER = 'ARRAY_BUFFER',
+	ELEMENT_ARRAY_BUFFER = 'ELEMENT_ARRAY_BUFFER',
+	INVERSE_BIND_MATRICES = 'INVERSE_BIND_MATRICES',
+	OTHER = 'OTHER',
+}
 
 /**
  * Model class providing writing state to a {@link Writer} and its {@link Extension}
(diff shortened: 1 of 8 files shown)


donmccurdy:
@zeux quick status update — I have implemented support equivalent to gltfpack's `-c` mode, and am seeing very similar compression percentages. I'm still tracking down a few bugs, but they're as likely to be coming from my quantization implementation as anywhere else.

I haven't spent much time tuning the higher `-cc` / filter mode, and so at the moment I am getting larger output than without it. I'll keep working on that. The basic current filter assignments are here, any pointers would be welcome! 

https://github.com/donmccurdy/glTF-Transform/blob/84b24dd82f0e51ecaa3963c5957759b08e9ba6cc/packages/extensions/src/ext-meshopt-compression/meshopt-compression.ts#L484-L508

Aside – is it possible to compress the accessor for a linear animation sampler output targeting `weights`? I have an example stored as normalized Uint8Array, type=SCALAR, which doesn't play well with the minimum byteStride=4 requirement... So far that appears to be the only accessor type I need to leave completely uncompressed.

Current compression stats (comparing gltfpack, gltf-transform + meshopt, and draco, below:

https://docs.google.com/spreadsheets/d/1vjyJabN24kX8DsGdyp832MOmNc0_8uciy6MIXqW5XBc/edit?usp=sharing

zeux:
Nice! Great to see this is moving along and I'm happy I finally got to making meshopt available as a library :D

gltfpack leaves weight animation track uncompressed right now. It's an annoying edge case I discovered a bit late and there was not a great way to correct it:

- The %4 requirement for stride isn't arbitrary, a non-divisible-by-4 stride makes SIMD implementation very difficult
- Additionally, to get perfect compression ratio you need to choose stride to match the weight count, but of course this means that you need many buffer views with different strides when different counts of targets are used in different meshes

I forget why gltfpack doesn't do it but a better-than-nothing option could be to use byteStride=4 with an artificially selected count=size/4 (and pad the size to 4) on this stream. I can try it and report back, I don't think there are any restrictions on the glTF spec or encoding/decoding side that will make this invalid but I don't remember why I didn't do that.

zeux:
For NORMAL you should use 4-byte or 8-byte normalized signed integer quantization and octahedral filter

For rotation you should use 8 byte normalized signed integer quantization and quaternion filter.

zeux:
> I forget why gltfpack doesn't do it but a better-than-nothing option could be to use byteStride=4

It was probably a little bit inconvenient, but it's easy enough to implement. Unfortunately I have exactly one real-world file in my test library that uses morph target animations and the results aren't super conclusive:

-rwxrwxrwx 1 zeux zeux 29639932 Aug 13 20:46 izzy.glb           
-rwxrwxrwx 1 zeux zeux 28483231 Aug 13 20:46 izzy.glb.gz        
-rwxrwxrwx 1 zeux zeux 29608896 Aug 13 20:46 izzy_wcmp.glb      
-rwxrwxrwx 1 zeux zeux 28522537 Aug 13 20:46 izzy_wcmp.glb.gz   

-31 KB before gzip (the raw morph animation data here is ~120 KB), +39 KB after gzip

I wouldn't necessarily read a whole lot into whole-file gzip results because gzip tends to be very flaky on stream changes that change the size (the compressed size can vary by a few % from adding a single byte mid-stream in some cases), but without comprehensive data it's hard to declare this an obvious win.

If we get ability to emit single-weight animation tracks at some point (I feel like we discussed it in some glTF issue/PR but I can't find it quickly atm) then this technique can be very useful/powerful because the variable stride problem goes away and fixed stride=4 should work fine, but until then leaving these uncompressed is probably a safer bet.

donmccurdy:
Many thanks! I believe `-c` is working well now, after a number of fixes to my `quantize()` function. I'm continuing on support for a `-cc` mode in https://github.com/donmccurdy/glTF-Transform/pull/337. I have some other examples for morph target weight animation outputs, but haven't had the chance to look further into that yet, so they are currently left uncompressed here.

PR #326, opened 2021-08-12
Write accessor min/max conditionally

### LINKED ISSUE #325: Reduce accessor min/max size in JSON
Current output includes min/max for every accessor, and writes them with high precision:

```
      "max": [
        0.9957213401794434,
        0.9957212805747986,
        0.9957213401794434
      ],
      "min": [
        9.9999988758398e-13,
        9.9999988758398e-13,
        9.9999988758398e-13
      ],
```

This takes up more space than necessary in the JSON output. We could omit min/max for many accessor types, and truncate precision somewhat where we do include it.

donmccurdy:
Fixes #325.

### DIFF
--- a/packages/core/src/io/writer-context.ts
+++ b/packages/core/src/io/writer-context.ts
@@ -41,16 +41,16 @@ export class WriterContext {
 	public readonly accessorParents = new Map<Property, Set<Accessor>>();
 
 	constructor (
-		doc: Document,
+		private readonly _doc: Document,
 		public readonly jsonDoc: JSONDocument,
 		public readonly options: Required<WriterOptions>
 	) {
-		const root = doc.getRoot();
+		const root = _doc.getRoot();
 		const numBuffers = root.listBuffers().length;
 		const numImages = root.listTextures().length;
 		this.bufferURIGenerator = new UniqueURIGenerator(numBuffers > 1, options.basename);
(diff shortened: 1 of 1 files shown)


packages/core/src/io/writer-context.ts
```diff
@@ -111,9 +111,19 @@ export class WriterContext {
...
-		accessor.getMax((accessorDef.max = []));
-		accessor.getMin((accessorDef.min = []));
-		accessorDef.normalized = accessor.getNormalized();
+
+		const needsBounds = this._doc.getGraph()
+			.listParentLinks(accessor)
+			.some((link) => link.getName() === 'POSITION');
+		if (needsBounds) {
+			accessor.getMax((accessorDef.max = []).map(Math.fround));
+			accessor.getMin((accessorDef.min = []).map(Math.fround));
```
lexaknyazev, line 120: If `accessor.getMax` fetches values from binary buffers, then `Math.fround` is a no-op.

donmccurdy: Yes, missed this! Fixed in https://github.com/donmccurdy/glTF-Transform/commit/dedd34c1b78986a214b4730149f3c29799146bd4.

donmccurdy:
Cleanup + bug fix in https://github.com/donmccurdy/glTF-Transform/commit/dedd34c1b78986a214b4730149f3c29799146bd4.

PR #332, opened 2021-08-17
Add node and scene preread hooks

robertlong:
These preread hooks are necessary for the `MOZ_hubs_components` extension. The component doesn't have a stable schema, so I'd like to migrate from older versions to a singular up to date version by modifying the json in the preread hook.


donmccurdy:
Thanks!

PR #368, opened 2021-09-30
Corrected multi-byte units in format function

rawnsley:
SI units for multi-byte data require base-10 convention, not the base-2 convention, which should use the alternative prefixes such as kibi.

Ref: https://en.wikipedia.org/wiki/Kilobyte


donmccurdy:
Thanks!

PR #369, opened 2021-09-30
Don't try and compress unsupported image formats

rawnsley:
The `toktx` tool errors out if you try and compress anything but PNG and JPEG images. Although nothing else is allowed by the core GLTF spec it is common for third-party extensions to include other things, for instance _image/vnd.radiance_ for HDR lightmaps or environments. Downgrading this to a warning at least allows the image to remain unchanged in the output file.

This could also be addressed using the `slots` property, but it's a bit of a blunt tool and hard to use when you want to target specific image types.


donmccurdy:
Agreed, thank you!

PR #380, opened 2021-10-06
Add DenoIO

### LINKED ISSUE #379: Add DenoIO
If only for adventure's sake, I think it'd be interesting to try to have glTF-Transform support the Deno runtime (https://deno.land/). First step would presumably be a `DenoIO` class, then running the test suite in CI on Deno. 

I don't necessarily expect the entire `/functions` package to work out of the box, given WASM dependencies and such, but `/core` and `/extensions` should be pretty doable.

donmccurdy:
- Fixes #379

### DIFF
--- a/docs/theme/layouts/default.hbs
+++ b/docs/theme/layouts/default.hbs
@@ -96,6 +96,11 @@
 			</a>
 		</li>
 		<li class="tsd-kind-class">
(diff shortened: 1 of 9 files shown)


donmccurdy:
Example of Deno CI integration: https://github.com/actions/starter-workflows/blob/da223f8a03b845b8624b144879658659a0b97d76/ci/deno.yml

Issue #398, opened 2021-10-25
Material Dedup functionality


MrMagicPenguin:
When preparing a series of glTF objects for sequencing, the materials for each 'frame' object are kept despite being the same across all objects. This may be a side effect of the workflow I am using and creating a solution for (i.e. OBJ Sequence from Blender -> single glTF sequence animation)

**Describe the solution you'd like**
Similar to the existing [`dedup`](https://github.com/donmccurdy/glTF-Transform/blob/f2e37b9/packages/functions/src/dedup.ts#L31) function, take the base materials from a mesh and remove the extraneous duplicate materials, then reassign the original material to each subsequent 'frame' mesh.

**Describe alternatives you've considered**
The current workaround I am using is a naive(see: hamfisted) approach where I simply loop through all the materials on the first mesh in the first frame, ignore the first material then detatch & dispose the materials on subsequent meshes. Then copy the material from the first mesh to the subsequent meshes. It would be neat to have this ability packed in as an option for the dedup function as is. The second part of this is not present in the below gist, i havent gotten as far as to write both that functionality or a better test animation for it.

https://gist.github.com/MrMagicPenguin/db149a701f6e36fade56349884756238

If this meshes (pun intended) with your vision for gltf-transform but isnt reasonably worth your time i'd be happy to give writing it as an extension a shot since I will need to make this happen regardless.

Cheers!

MrMagicPenguin:
Sent a PR - Apologies in advance for my code formatting not matching your conventions - I recently switched to Linux and use Webstorm, and I haven't got all the kinks out just yet.

Please let me know if this works for you, and I'll get cracking on the `dedup()` ability. The one thing I can think of is handling the 'extras' correctly, but I am not 100% sure how to address that yet.

PR #405, opened 2021-10-30
Add .equals() method to Material property

MrMagicPenguin:
Add method for [`Material`](https://gltf-transform.donmccurdy.com/classes/core.material.html) to compare material instances. Useful for checking duplicate material; planned as first step for #398.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ node_modules
 .nyc_output
 .rpt2_cache
 .vscode
+.idea
 **/out/*
 !**/out/.gitkeep
 packages/**/dist/*
--- a/docs/CLI_HELP.md
+++ b/docs/CLI_HELP.md
@@ -1,5 +1,5 @@
 
-  gltf-transform 1.0.0 — Commandline interface for the glTF-Transform SDK.
+  gltf-transform 1.0.1 — Commandline interface for the glTF-Transform SDK.
 
   USAGE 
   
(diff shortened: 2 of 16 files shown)


donmccurdy:
Thank you! I've merged this in https://github.com/donmccurdy/glTF-Transform/commit/e27344afdb77a160eac0961d8cb39e47e8292259, and made a few adjustments in https://github.com/donmccurdy/glTF-Transform/commit/7db0ffd287c7b58a94dd39c9aa189a99c085c5ea. For the other PR (#407) you may need to resolve conflicts or rebase the branch to get synced with the latest 'main' branch.

.gitignore
```diff
@@ -7,6 +7,7 @@ node_modules
 .nyc_output
 .rpt2_cache
 .vscode
+.idea
```
donmccurdy, line 10: I omitted this line, but FWIW you can put this into a `.gitignore` file in your user directory (e.g. `~/.gitignore`) and it should apply to all projects you work on. This makes it easier to contribute to multiple projects without having to customize `.gitignore` for each.

MrMagicPenguin: Hey thanks for the tip!! I should have simply omitted this change, but I will keep that in mind for the future. Cheers!

z1z2:
@donmccurdy If one material has baseColorTexture and another hasn't, and other properties of the two materials are the same, the equals method will still return true.

```ts
		if (this.getBaseColorTexture() && other.getBaseColorTexture()) {
			if (this.getBaseColorTexture() !== other.getBaseColorTexture()) return false;
			if (!this.getBaseColorTextureInfo()!.equals(other.getBaseColorTextureInfo()!)) {
				return false;
			}
		}
```

Issue #414, opened 2021-11-04
[CLI] meshopt compression breaks UV coordinates


hybridherbst:
**Describe the bug**
UV coordinates seem to break in some cases when applying meshopt compression.

**To Reproduce**

1. Download the attached file [MultipleUVRanges.zip](https://github.com/donmccurdy/glTF-Transform/files/7477635/MultipleUVRanges.zip)
2. Run ```gltf-transform meshopt MultipleUVRanges.glb MultipleUVRanges.meshopt.glb```
3. Open the file e.g. in https://sandbox.babylonjs.com/ (doesn't load in gltf.report)
4. Note distorted UV coordinates on output file
![WRYJMRK14U](https://user-images.githubusercontent.com/2693840/140405817-85f21e01-1e73-4f5f-ae6c-49696e265d60.png)

**Expected behavior**
Output file looks the same as input file
![chrome_ZiTdOLGFe9](https://user-images.githubusercontent.com/2693840/140405802-0551bf56-68be-4dde-b420-982cd56b4a28.png)

**Versions:**
 - Version: 1.0.1
 - Environment: nodejs

**Notes**
- doesn't seem to happen when running ```gltfpack -c``` or ```-cc``` on the same file

donmccurdy:
Likely the same as https://github.com/donmccurdy/glTF-Transform/issues/403, but I'm glad for the extra test case thanks! It seems like the issue shows up when using `--level high` (similar to `-cc` in meshopt) but not `--level medium` (similar to `-c`).

donmccurdy:
Just curious, what are the "Multiple UV ranges" referred to by the title?

donmccurdy:
> I thought it may be because of those UV ranges (and gltf-transform logging that UVs out of [0..1] range are skipped) but seems I'll have to hunt some more...

The error doesn't show up when using just the `quantize` command (which is always a pre-process step when doing meshopt) so those warnings shouldn't be the issue. I think given that the error only shows up with the higher meshopt compression setting, I may be using one of the meshopt compression filters incorrectly here. I'll take a look!

donmccurdy:
Merging details from #403 —

***


Testing [War Plane (Sketchfab)](https://sketchfab.com/3d-models/war-plane-b487f1128b984da8ae8cd9ce0a02d9ba) on https://gltf.report, the model looks OK when exported normally, or when quantized, but enabling Meshopt compression results in broken UVs. Test on CLI as well, which has the same error.

[...]

PR #420, opened 2021-11-09
Avoid meshopt compression on unquantized UVs

donmccurdy:
Intended to work around #414, but doesn't seem to be working yet.


packages/extensions/src/ext-meshopt-compression/meshopt-compression.ts
```diff
@@ -277,6 +278,14 @@ export class MeshoptCompression extends Extension {
 			// Example: https://skfb.ly/6qAD8
 			if (getTargetPath(accessor) === 'weights') continue;
 
+			// See: https://github.com/donmccurdy/glTF-Transform/issues/414
+			const isTexcoord = listSemantics(accessor, this.doc)
+				.some((semantic) => semantic.startsWith('TEXCOORD_'));
```
donmccurdy, line 283: If I omit this TEXCOORD_0 check and skip the unquantized NORMAL and TANGENT properties as well the output is fine... why do we need to skip more than TEXCOORD_*? And why are NORMAL and TANGENT not quantized?

donmccurdy:
See #414.

PR #431, opened 2021-11-19
Add dequantize() function

donmccurdy:
Adds `dequantize()` function, removing `KHR_mesh_quantization` and converting vertex attributes to `float32`.

Usage:

```js
import { dequantize } from '@gltf-transform/functions';
await document.transform(dequantize());
```

```shell
gltf-transform dequantize input.glb output.glb
```

Remaining:

- [x] Unit tests

### DIFF
--- a/packages/functions/src/dequantize.ts
+++ b/packages/functions/src/dequantize.ts
@@ -8,13 +8,13 @@ const NAME = 'dequantize';
 export interface DequantizeOptions {
 	/**
 	 * Pattern (regex) used to filter vertex attribute semantics for quantization.
-	 * Default: `/^(JOINTS)(_\d+)$/`.
+	 * Default: `/^((?!JOINTS)(_\d+))*$/`.
 	 */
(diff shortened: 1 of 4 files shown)


packages/functions/src/dequantize.ts
```diff
@@ -0,0 +1,65 @@
...
+export interface DequantizeOptions {
+	/**
+	 * Pattern (regex) used to filter vertex attribute semantics for quantization.
+	 * Default: `/^(JOINTS)(_\d+)$/`.
+	 */
+	pattern?: RegExp;
+}
+
+const DEQUANTIZE_DEFAULTS: DequantizeOptions = {
+	pattern: /^(JOINTS)(_\d+)$/,
```
donmccurdy, line 17: Typo – this should be _excluding_ `JOINTS_*`, not including it. Presumably my manual tests worked because the CLI default is correct, but this should fail in unit tests.

Issue #452, opened 2021-12-13
Support remote URLs in Node.js and CLI


daniel-gato:
As per the specifications external URLs for assets are not forbidden in gltfs for textures and .bins.

Somehow, the cli utility is trying to fetch the url from the current path: 

```bash
error: ENOENT: no such file or directory, open '/Users/daniel/Desktop/https:/cdn.exmaple.io/dvpt-dg/shared_assets/files/000/001/465/9e3f8087b67dcaf4d1bed33bd23699b760637e6bf2e3a2c7883dc2e6933c9e04-optimised_jpg.jpg'
```

Am I missing something here?

I understand that `validate` would say :
```
Non-relative URI found: 'https://....jpg'.glTF Validator(NON_RELATIVE_URI)
```

But this shouldn't block the other cli commands.

Any thoughts?

donmccurdy:
Currently this project has two I/O classes: NodeIO and WebIO. NodeIO doesn't make network requests (you generally don't want that in a server environment), and so the CLI doesn't make network requests either. I'm not completely sure the CLI should make network requests silently either, but certainly an opt-in flag (e.g. `--allow-remote`) would be a valid feature request. And in the meantime it should probably show a clearer error rather than trying to resolve the URL on disk. So I'll mark this as a feature request for now.

One option, in the meantime, is to do processing through the scripting API instead. For example:

```js
const path = require('path');
const fetch = require('node-fetch');

const json = await fetch('path/to/model.gltf').then((res) => res.json());
const resources = {};

for (const image in json.images) {
  const basename = path.basename(image.uri);
  resources[basename] = await fetch(image.uri);
  image.uri = basename;
}
for (const buffer in json.buffers) {
  ...
}

const document = new NodeIO().readJSON({json, resources});
```

HexaField:
Hi there! I would love to support this issue. The XREngine team already uses this library and we are stuck not being able to load instanced GLTF models on our nodejs backend.

I believe there is a significant use case for loading GLTFs on a nodejs server, for things like user generated content, dynamic scene loading, server side 3d rendering & pixel streaming, various NFT functionality, the list goes on...

Please let us know how you think this could be best implemented and we will make swift work of it. I believe all that is needed is for the ability to pass a `URL` object into `fs.readFileSync` instead of a uri string, though I understand this may be more complicated when thinking about .gltf with internally referenced relative .bin paths.

donmccurdy:
[...]

PR #462, opened 2021-12-28
Add Node IO option for using https fetch

HexaField:
Adds an initial implemention regarding the discussion https://github.com/donmccurdy/glTF-Transform/issues/452

This is just to ensure the solution is feasible for our use case. Now that it is working I will do some refactoring to make it more along the lines of what has been suggested in the linked issue.

Checklist
- [x] initial implementation
- [x] receive feedback
- [x] write appropriate tests
- [x] write documentation & issue changelog

This contribution is funded by [XRFoundation](https://xrfoundation.io).


packages/core/src/io/node-io.ts
```diff
@@ -34,31 +34,49 @@ import { PlatformIO } from './platform-io';
...
 	private _fs;
 	private _path;
+	private _fetch;
+	private _httpRegex = /https?:\/\//;
+	public useFetch = false;
 
 	/** Constructs a new NodeIO service. Instances are reusable. */
-	constructor() {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	constructor(fetch?: any) {
```
donmccurdy, line 43: minor: Using `unknown` instead of `any` here should allow you to avoid `eslint-disable-next-line`.

packages/core/src/io/node-io.ts
```diff
@@ -34,31 +34,49 @@ import { PlatformIO } from './platform-io';
...
+    this.useFetch = !!fetch;
 	}
 
 	protected async readURI(uri: string, type: 'view'): Promise<Uint8Array>;
 	protected async readURI(uri: string, type: 'text'): Promise<string>;
 	protected async readURI(uri: string, type: 'view' | 'text'): Promise<Uint8Array | string> {
 		switch (type) {
-			case 'view':
+			case 'view': {
+        if(this.useFetch && this._httpRegex.exec(uri)) {
```
donmccurdy, line 57: minor: Let's check httpRegex first, and if it matches but `useFetch` is disabled we can throw a clear error. Avoids trying to read a bad path on disk like #452 and tells the user how to fix it.

packages/core/src/io/node-io.ts
```diff
@@ -34,31 +34,49 @@ import { PlatformIO } from './platform-io';
...
+				} 
 				return this._fs.readFile(uri, 'utf8');
+      }
 		}
 	}
 
 	protected resolve(directory: string, path: string): string {
-		return this._path.resolve(directory, path);
+    if(this.useFetch) return `${directory}/${path}`;
+		return this._path.join(directory, path);
```
donmccurdy, line 75: I can't quite tell from the CI output why the tests are failing, but I think it might be this method. One important difference between `path.resolve` and `path.join` is that `resolve` will ignore the directory if the path is already absolute. Perhaps the logic here should be:

1. If `directory` or `path` matches httpRegex, use the [same resolver as WebIO](https://github.com/donmccurdy/glTF-Transform/blob/290e028542e1c23fe10c4f940a84b6181a5e3b72/packages/core/src/io/web-io.ts#L68-L87)
2. If neither matches httpRegex, use `_path.resolve` as before.

In retrospect, probably `basePath` or `rootPath` would have been better names than `directory` here.

packages/core/src/io/node-io.ts
```diff
@@ -34,31 +34,49 @@ import { PlatformIO } from './platform-io';
...
 	}
 
 	protected dirname(uri: string): string {
+    if(this.useFetch) return uri.split('/').slice(0, -1).join('/');
```
donmccurdy, line 79: See above – probably best to test against `httpRegex` first here.

donmccurdy (commented):
Thank you! These changes are looking good, and I've added a few suggestions.

Do you know whether the changes are working for your use case so far?

donmccurdy:
One (very optional) note – if you install [Prettier](https://prettier.io/) for your IDE it should automatically format and indent everything. No problem if you'd rather not, though, it will just get auto-formatted the next time I touch the file.

packages/core/src/io/node-io.ts
HexaField: Using unknown then causes this._fetch to then be of type unknown, thus causing errors where it is used.

HexaField:
> One (very optional) note – if you install [Prettier](https://prettier.io/) for your IDE it should automatically format and indent everything. No problem if you'd rather not, though, it will just get auto-formatted the next time I touch the file.

Ah thanks, I'm used to pre-commit hooks doing that for me.

HexaField:
Implemented the feedback provided added some tests, though it seems there is an error occuring I am struggling with. `Error: Missing required GLB JSON chunk.` shows up when the gltf read test runs for http files.

packages/core/src/io/node-io.ts
donmccurdy: This just requires casting to a known type, see [this example](https://github.com/donmccurdy/glTF-Transform/blob/462cd759471d5b65df8d5c54379a09e139785a48/packages/core/src/io/deno-io.ts#L37-L42). I can adjust the TS types after this is merged too though.

donmccurdy:
I think that error originates from the new tests, but doesn't fire until later when some other test is running. You can run a single test in isolation with `test.only`:

```js
test.only('@gltf-transform/core::io | node.js read glb http', async (t) => { ... });
```

I'll leave a few more comments inline.

packages/core/test/io/node-io.test.ts
```diff
@@ -44,6 +55,36 @@ test('@gltf-transform/core::io | node.js read gltf', { skip: environment !== Env
 	t.end();
 });
 
+test('@gltf-transform/core::io | node.js read glb http', { skip: environment !== Environment.NODE }, async (t) => {
+	let count = 0;
+	glob.sync(path.join(__dirname, '../in/**/*.glb')).forEach((inputURI) => {
+		const basepath = inputURI.replace(path.join(__dirname, '../in'), 'http://' + ip);
+
+		const io = new NodeIO(require('node-fetch'));
+		const doc = io.read(basepath);
```
donmccurdy, line 64: Because `io.read` is now asynchronous, it return a promise — this might be throwing errors later, after the test has completed and purportedly passed. One pattern you could use here would be:

```js
	const uris = glob.sync(path.join(__dirname, '../in/**/*.glb'));
	await Promise.all(
		uris.map(async (inputURI) => {
			const io = new NodeIO();
			const doc = await io.read(inputURI);
            // ...
		})
	);
```

packages/core/test/io/node-io.test.ts
```diff
@@ -2,11 +2,22 @@ import test from 'tape';
...
+
+	server = http.createServer((req, res) => {
+		const filePath = path.join(path.join(__dirname, '../in'), req.url);
+		const file = fs.readFileSync(filePath, 'utf-8');
+		res.write(file);
+		res.end();
+	});
+	
+	server.listen(8989);
 }
```
donmccurdy, line 21: Running tests locally I'm seeing the following error:

```
error: request to http://192.168.1.124:8989/Box_glTF-Binary/Box.glb failed, reason: connect ECONNREFUSED 192.168.1.124:8989
(node:69294) UnhandledPromiseRejectionWarning: FetchError: request to http://192.168.1.124:8989/BoxVertexColors.glb failed, reason: connect ECONNREFUSED 192.168.1.124:8989
    at ClientRequest.<anonymous> (/Users/donmccurdy/Documents/Projects/glTF-Transform/node_modules/node-fetch/lib/index.js:1461:11)
    at ClientRequest.emit (events.js:400:28)
    at ClientRequest.emit (domain.js:475:12)
    at Socket.socketErrorListener (_http_client.js:475:9)
    at Socket.emit (events.js:400:28)
    at Socket.emit (domain.js:475:12)
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:82:21)
```

Perhaps it would be easier just to 'mock' the fetch API? See how tests for WebIO do this:

https://github.com/donmccurdy/glTF-Transform/blob/6b9f4791a7c5cea1aa42674dfdf4055098a06833/packages/core/test/io/web-io.test.ts#L11-L30

In this case we wouldn't have to patch the global namespace since `fetch` is already passed as a constructor parameter.

donmccurdy:
Thanks, tests are passing now! Is this ready for final review and then merging, in your opinion?

HexaField:
> Thanks, tests are passing now! Is this ready for final review and then merging, in your opinion?

Yes I believe so, I can make follow up PRs if need be after some more testing. Thanks!

donmccurdy:
Awesome, thank you @HexaField! I've merged the PR, but am also planning to require a `setAllowHTTP(bool)` method to make this a bit more explicit. See https://github.com/donmccurdy/glTF-Transform/pull/466.

HexaField:
Thank you!

PR #468, opened 2022-01-04
Add KHR_xmp_json_ld

### LINKED ISSUE #65: Add 'xmp' lib and CLI function
https://github.com/KhronosGroup/glTF/pull/1553

donmccurdy:
Changes:
- Fixes #65

Summary:

Provides support for reading/writing XMP metadata using the [KHR_xmp_json_ld](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_xmp_json_ld/) extension. On the commandline, glTF-Transform will offer three methods of updating XMP metadata: interactive prompts, external JSONLD files, and automatic updates to certain timestamp fields like `xmp:ModifyDate`. Users requiring greater control over the process — or wishing to incorporate XMP metadata into their web application and/or server workflows — may use the scripting API to manually configure any metadata field.

CLI Usage:

- Read XMP:
    - `gltf-transform inspect input.glb`
- Embed XMP with interactive prompts:
    - `gltf-transform xmp input.glb output.glb`
- Embed XMP with external .jsonld:
    - `gltf-transform xmp input.glb output.glb --packet metadata.jsonld`
- Reset XMP fields:
    - `gltf-transform xmp input.glb output.glb --reset`

Scripting usage:

```ts
import { XMP, Packet } from '@gltf-transform/extensions';

// Create an Extension attached to the Document.
const xmpExtension = document.createExtension(XMP);

// Create Packet property.
const packet = xmpExtension.createPacket()
    .setProperty('dc:Creator', {'@list': ['Acme, Inc.']});

// Option 1: Assign to Document Root.
document.getRoot().setExtension('KHR_xmp_json_ld', packet);

// Option 2: Assign to a specific Property.
texture.setExtension('KHR_xmp_json_ld', packet);
```

Documentation: _(pre-release, temporary URLs)_

- XMP: `https://gltf-transform-donmccurdy.vercel.app/classes/extensions.xmp.html`
- Packet: `https://gltf-transform-donmccurdy.vercel.app/classes/extensions.packet.html`

Remaining tasks:
- [x] ~Support XMP packets attached to specific resources (not just `root.asset`) in scripting API~
- [x] ~Maintain `xmp:ModifyDate` and `xmp:MetadataDate` automatically~
- [x] ~Unit tests~
- [x] Documentation
- [ ] (Optional) Pre-populate existing data for interactive mode
- [ ] (Optional) Generate per-scene (?) XMP packets in `merge` command

_This contribution is funded by the [Khronos Group](https://www.khronos.org/)._


donmccurdy:
Open question: Is it safe to search the JSON keys for `term:*` prefixes, so that we can automatically include common context terms on the users behalf, without putting unnecessary context terms into the JSON LD output? Need to know whether there are any cases where a context term would be necessary, but _not_ included in any JSON key. For example, `{"@type": "rdf:Alt"}` is valid, but always followed by `"rdf:_1": {...}` so we'll still find a key.

If there isn't a robust way to detect which context is being used (for arbitrary JSON LD input), I'll need to require users to specify that explicitly:

Before:

```ts
packet.setProperty('dc:Creator', {'@list': ['Acme, Inc.']});
```

After:

```ts
packet
  .setContext('dc', 'http://purl.org/dc/elements/1.1/')
  .setProperty('dc:Creator', {'@list': ['Acme, Inc.']});
```

weegeekps:
> ... unless it is possible to robustly detect which JSONLD context terms are needed, I think I will need to require users to specify them for each packet. Context terms other than the common ones (dc, rdf, xmp, xmpRights, model3d) would need to be specified in any case.

I went back and looked through how a couple other libraries are handling JSON-LD contexts and honestly I think the idea of having a `setContext` method that must be called on the packet is the sanest solution, especially in the context of XMP data. A lot of competing approaches require expanded URIs which is something we discourage with `KHR_xmp_json_ld`.

A user setting metadata should know the namespace and namespace URI as they're setting the data, so requiring them to be specified seems reasonable.

donmccurdy:
Thanks! Going that route sounds good to me.

donmccurdy:
Merging this PR. Some remaining optional additions are described in https://github.com/donmccurdy/glTF-Transform/issues/476.

donmccurdy:
An alpha release including this PR has been published under the `@next` tag. For example:

```
npm install --global @gltf-transform/cli@next
```

Issue #489, opened 2022-02-08
Animation compression gets worse over time when using meshopt


hybridherbst:
**Describe the bug**
It seems that gltf-transform meshopt currently has "medium" and "high" presets, but it looks like in both these presets animations degrade more the longer the animation time is (keyframes that are later in time compress worse).

**To Reproduce**
1. Download [AnimatedArm.zip](https://github.com/donmccurdy/glTF-Transform/files/8023253/AnimatedArm.zip)
2. This contains Arm.glb, Arm.meshopt-medium.glb and Arm.meshopt-high.glb
2. The effect is barely visible on medium but very visible on high; it seems that the further to the back of an animation the values are, the worse the compression ends up being.

**Expected behavior**
My understanding is that the error/lossy compression should be uniformly distributed and not worse to the "back" of the animation, but I might be mistaken in my meshopt animation compression understanding...

**Additional context**
Babylon Sandbox works good for viewing the files as it has a timeline slider (https://sandbox.babylonjs.com/).

The files are also good stress test files for https://gltf.report by the way, opening the "Validate" view crashes Chrome (out of memory) 👀

**EDIT:** I'm not sure if this is the same or a separate issue, but `Arm.meshopt-high.glb` also isn't valid anymore since it contains duplicate time values (ACCESSOR_ANIMATION_INPUT_NON_INCREASING), and, a lot of them, which is what I think is crashing gltf.report.

hybridherbst:
When looking at the imported animation it seems that indeed the time itself is nonlinearly packed (time resolution becomes lower over time) which I think might be a bug:
![image](https://user-images.githubusercontent.com/2693840/152983588-00a23f1e-07b6-4ec6-936f-296de975a482.png)

donmccurdy:
Animation times and values can use different filter methods, but the other two filter types aren't appropriate for time. So if exponential doesn't work here I'd just need to disable the filters on animation times probably.

PR #492, opened 2022-02-09
Don't use meshopt filters for animation sampler input

donmccurdy:
Fixes https://github.com/donmccurdy/glTF-Transform/issues/489.

@hybridherbst this passes validation, but could you confirm the asset created after this patch (attached) has the animation data you expect? I'm not sure how to reproduce the graph you showed in the original issue. Thanks! 

[Arm.CUST-meshopt-high.glb.zip](https://github.com/donmccurdy/glTF-Transform/files/8028460/Arm.CUST-meshopt-high.glb.zip)

### DIFF
--- a/packages/extensions/src/ext-meshopt-compression/encoder.ts
+++ b/packages/extensions/src/ext-meshopt-compression/encoder.ts
@@ -172,7 +172,7 @@ export function getMeshoptFilter(accessor: Accessor, doc: Document): { filter: M
 		}
 
 		// See: https://github.com/donmccurdy/glTF-Transform/issues/489
-		if (refName === 'input') return { filter: MeshoptFilter.NONE, bits: 12 };
+		if (refName === 'input') return { filter: MeshoptFilter.NONE };
 
 		if (refName === 'inverseBindMatrices') return { filter: MeshoptFilter.NONE };
 	}
--- a/packages/extensions/src/ext-meshopt-compression/encoder.ts
+++ b/packages/extensions/src/ext-meshopt-compression/encoder.ts
@@ -170,7 +170,10 @@ export function getMeshoptFilter(accessor: Accessor, doc: Document): { filter: M
 			if (targetPath === 'scale') return { filter: MeshoptFilter.EXPONENTIAL, bits: 12 };
 			return { filter: MeshoptFilter.NONE };
 		}
-		if (refName === 'input') return { filter: MeshoptFilter.EXPONENTIAL, bits: 12 };
+
+		// See: https://github.com/donmccurdy/glTF-Transform/issues/489
+		if (refName === 'input') return { filter: MeshoptFilter.NONE, bits: 12 };
+
 		if (refName === 'inverseBindMatrices') return { filter: MeshoptFilter.NONE };
 	}
 


hybridherbst:
From a quick glance it certainly looks like its much better distributed!
image

There's some increase in file size (595kB to 661kB) which I wouldn't quite have expected, but maybe besides a different keyframe distribution this actually results in more keyframes that are stored.

For reference, this is Unity's Animation Window, steps to look at a GLB animation would be:
- make sure you have the glTFast package installed in a Unity project, and potentially the meshopt/draco/ktx2 packages as needed
- drop some GLB file in Assets
- it'll import an animation clip (or multiple) if its animated
- double-click the animation clip
- click on "Curves"

Only reason I'm looking there is that there doesn't seem to be a good tool to look at glTF animation curves that is even remotely performant or usable (e.g. babylon kinda somewhat has that but it's not really usable if you have > 5 tracks or > 20 keyframes).

zeux:
FWIW the sizes are a little bit surprising here, but I haven't checked how glTF-Transform handles animation compression. Attaching the results for the relevant file with gltfpack with default settings, using `-c` and `-cc`. I'd be curious to know if the compressed animation here has observable issues, as it's difficult for me to tell visually. The -cc option produces a 163 KiB file (43 KiB after gzip compression).

[arm-gltfpack.zip](https://github.com/donmccurdy/glTF-Transform/files/8077012/arm-gltfpack.zip)

donmccurdy:
glTF Transform optimizations are split into separate operations, so it may require a few steps to get anywhere close to gltfpack. In this case it looks like gltfpack is resampling animation automatically, whereas glTF Transform doesn't automatically do that when applying meshopt.

Comparing _uncompressed_ animation data in various outputs — different channel, sampler, and keyframe counts in the output are probably the source of the size difference:

**original:**

| #   | name                         | channels | samplers | duration | keyframes | size     |
| --- | ---                          | ---      | ---      | ---      | ---       | ---      |
| 0   | Recorded_20220208-121134.glb | 63       | 63       | 337.741  | 759,501   | 11.14 MB |

**gltf-transform meshopt:**

| #   | name                         | channels | samplers | duration | keyframes | size    |
| --- | ---                          | ---      | ---      | ---      | ---       | ---     |
| 0   | Recorded_20220208-121134.glb | 63       | 63       | 337.741  | 759,501   | 5.87 MB |

**gltf-transform resample + dedup + meshopt:**

| #   | name                         | channels | samplers | duration | keyframes | size    |
| --- | ---                          | ---      | ---      | ---      | ---       | ---     |
| 0   | Recorded_20220208-121134.glb | 63       | 63       | 337.741  | 303,876   | 2.84 MB |

**gltfpack -cc**

| #   | name                         | channels | samplers | duration | keyframes | size     |
| --- | ---                          | ---      | ---      | ---      | ---       | ---      |
| 0   | Recorded_20220208-121134.glb | 6        | 6        | 337.767  | 60,804    | 567.5 KB |

I can't find any range of the `resample` command's `--tolerance` flag that gives results as good as gltfpack here, and the default output is quite broken, so I think there might be issues that need more work in the resampling implementation. I haven't looked into why the channel and sampler counts are so different yet.

Links:

- https://github.com/donmccurdy/glTF-Transform/issues/501
- https://github.com/donmccurdy/glTF-Transform/blob/ca187965eec975fc7027d473be39d84b0cff5b29/packages/functions/src/resample.ts#L72-L142

PR #389, opened 2021-10-19
Parallel texture compression, and recover if texture compression fails

mikejurka:
Parallelize compression (~4x faster for some models with many textures)

Uses p-limit for a concurrency pool
Read files async when doing texture compression.
Make the number of concurrent toktx jobs a command-line argument.
If a texture in a model fails to compress, don’t bail.

PR #539, opened 2022-04-06
toktx: Adjust jobs/threads count, add prefixes to 'toktx' log output

donmccurdy:
Follow-up to https://github.com/donmccurdy/glTF-Transform/pull/389, part 2 / 2. Apply suggested default jobs and threads counts. Adds prefixes to logs so they can be traced back to texture indices.

### DIFF
--- a/packages/cli/src/transforms/toktx.ts
+++ b/packages/cli/src/transforms/toktx.ts
@@ -12,6 +12,7 @@ import { spawn, commandExists, formatBytes, getTextureChannels, getTextureSlots,
 
 tmp.setGracefulCleanup();
 
+const NUM_CPUS = os.cpus().length || 1; // microsoft/vscode#112122
 const KTX_SOFTWARE_VERSION_MIN = '4.0.0-rc1';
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/transforms/toktx.ts
```diff
@@ -125,28 +130,29 @@ export const toktx = function (options: ETC1SOptions | UASTCOptions): Transform
 					texture.getURI() ||
 					texture.getName() ||
 					`${textureIndex + 1}/${doc.getRoot().listTextures().length}`;
-				logger.debug(`Texture ${textureLabel} (${slots.join(', ')})`);
+				const prefix = `toktx:texture(${textureLabel})`;
+				logger.debug(`${prefix}: Slots → ${slots.join(', ')}`);
```
donmccurdy, line 134: Looks better with brackets.

```suggestion
				logger.debug(`${prefix}: Slots → [${slots.join(', ')}]`);
```

mikejurka:
LGTM thanks!

PR #542, opened 2022-04-07
Skip compressing textures that are 4x4 or smaller.

mikejurka:
Was seeing texture turn green when compressing 1x1 textures with etc1s

### DIFF
--- a/packages/cli/src/transforms/toktx.ts
+++ b/packages/cli/src/transforms/toktx.ts
@@ -155,6 +155,10 @@ export const toktx = function (options: ETC1SOptions | UASTCOptions): Transform 
 					logger.warn(`${prefix}: Skipping, unreadable texture.`);
 					return;
 				}
+				if (size[0] <= 4 && size[1] <= 4) {
+					logger.warn(`${prefix}: Skipping, texture is 4x4 or smaller.`);
+					return;
+				}
 
 				// PREPARE: Create temporary in/out paths for the 'toktx' CLI tool, and determine
 				// necessary commandline flags.


donmccurdy:
Hm, there's also code later that will scale 5x5 textures down to 4x4:

https://github.com/donmccurdy/glTF-Transform/blob/1e6765d320e3edcd50e830c186c91f40cab38eb4/packages/cli/src/transforms/toktx.ts#L359-L367

I'm curious if you see the same issue when running that asset through gltfpack with the `-tc` flag? The current resizing logic is unfortunately already non-trivial:

https://github.com/donmccurdy/glTF-Transform/blob/1e6765d320e3edcd50e830c186c91f40cab38eb4/packages/cli/src/transforms/toktx.ts#L292-L320

It seems like something (gltf-transform? KTX-Software? the user?) should be responsible for scaling images to the right sizes for compression, at the moment that is on gltf-transform but I have no real insight into why a 4x4 texture wouldn't compress properly.

@lexaknyazev does this seem like something that should be reported as a KTX-Software issue? Or maybe documentation in KTX-Software or https://github.com/KhronosGroup/3D-Formats-Guidelines about image resolution expectations with these formats?

lexaknyazev:
There's nothing in the codec that would unconditionally turn a 4x4 texture green.

@mikejurka
Would you be able to share an example image file?

mikejurka:
> There's nothing in the codec that would unconditionally turn a 4x4 texture green.
> 
> @mikejurka Would you be able to share an example image file?

Sure! Download this model:
https://d1a370nemizbjq.cloudfront.net/66316cff-e2eb-4679-839f-abba0764c09f.glb

Then compress like this:
`gltf-transform etc1s 66316cff-e2eb-4679-839f-abba0764c09f.glb  output.glb`
Screen Shot 2022-04-07 at 12 42 40 PM

open in gltf.report and observe the arms are green. with this patch, it's fine.

lexaknyazev:
There are several points here.

- The original texture is 1x1, i.e., it represents a solid color. Generally, such textures should be avoided since a solid color can be represented via JSON-stored values, e.g., `baseColorFactor`.
- The JPEG file contains an embedded color profile that will be ignored by glTF viewers anyway. It takes 554 out of 842 bytes.
- The image turns green because of a bug in the Basis Universal encoder that seems to be already fixed. The issue is not reproducible with `toktx` built from the current head.

mikejurka:
Thanks! This isn't my model, we are generally in a position of handling user-submitted content.

I'll test with an up-to-date toktx, and this change is also unnecessary then.

mikejurka:
Unfortunately it looks like KTX hasn't released any pre-compiled version in the last year, but of course i can always compile it myself. Or am i looking in the wrong spot?
https://github.com/KhronosGroup/KTX-Software/releases

donmccurdy:
Thanks @lexaknyazev and @mikejurka — I'll leave the current resizing logic alone for now in that case. I'm not sure about the KTX-Software release status either unfortunately. I'm hoping https://github.com/GoogleChromeLabs/squoosh/pull/1017 might land at some point, at which point we could evaluate that tool.

PR #599, opened 2022-06-16
The node of IAnimationChannelTarget should be optional

06wj:
See https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-animation-channel-target
image


donmccurdy:
Thank you! Just to check -- does this change fix an issue affecting this library? or correct the type definitions for general use?

06wj:
My glTF file has extension for joint node, see the pic below.
image

There is a problem when manually creating the gltf JSON.

donmccurdy:
Ok I see - thank you!

PR #600, opened 2022-06-18
Add weight normalization option to quantize function

snagy:
Added an option to the quantize function that controls whether weights get normalized.  Default is true, so the behavior remains as before.

I have a private extension which relies on non-normal weights and I'd like to quantize it!


donmccurdy:
Thank you!

PR #622, opened 2022-07-14
Add texture slot filter to texture resize function

snagy:
Added a texture slot filter to the texture resize function.  This is similar to the sqoosh implementation - it takes a regex and passes if any of the texture slots match the regex.  If the slots value is null then all textures pass.


packages/functions/src/texture-resize.ts
```diff
@@ -2,6 +2,7 @@ import ndarray from 'ndarray';
 import { lanczos2, lanczos3 } from 'ndarray-lanczos';
 import { getPixels, savePixels } from 'ndarray-pixels';
 import type { Document, Transform, vec2 } from '@gltf-transform/core';
+import { listTextureSlots } from '@gltf-transform/functions';
```
donmccurdy, line 5:
```suggestion
import { listTextureSlots } from './list-texture-slots';
```

Minor — I think this may need to be a relative import since the target is in the same package.

donmccurdy (commented):
This looks great, thank you! One suggestion inline.

packages/functions/src/texture-resize.ts
snagy: You're right; surprisingly both seem to work.  Let's go with the relative version.

PR #663, opened 2022-08-17
Reorder dedupe property types

robertlong:
By deduping meshes after textures and materials you're more likely to be able to deduplicate more meshes.

### DIFF
--- a/packages/functions/src/dedup.ts
+++ b/packages/functions/src/dedup.ts
@@ -55,9 +55,9 @@ export const dedup = function (_options: DedupOptions = DEDUP_DEFAULTS): Transfo
 		const logger = doc.getLogger();
 
 		if (propertyTypes.has(PropertyType.ACCESSOR)) dedupAccessors(logger, doc);
-		if (propertyTypes.has(PropertyType.MESH)) dedupMeshes(logger, doc);
 		if (propertyTypes.has(PropertyType.TEXTURE)) dedupImages(logger, doc);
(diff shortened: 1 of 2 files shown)


donmccurdy:
Thanks @robertlong! I think the unit tests for this feature predate the addition of material deduplication (in https://github.com/donmccurdy/glTF-Transform/commit/f69517b8fd30a05ddc728ad5b28c5d59ff16c5c6), and the unit test expects mesh deduplication to fail for two materials that differ only in name:

https://github.com/donmccurdy/glTF-Transform/blob/0ffa610a82594df3394b3f0aed4d2cb7dce5d2bf/packages/functions/test/dedup.test.ts#L82-L100

Do mind updating the test to ensure that (1) meshes sharing materials that differ only in name WILL be deduplicated, and (2) meshes sharing materials that are meaningfully different WILL NOT be deduplicated?

robertlong:
Sounds good, I'm also noticing that there's a bug de-duping materials+meshes in Third Room right now so I'm looking into that first

robertlong:
Sounds good, I'm also noticing that there's a bug de-duping materials+meshes in Third Room right now so I'm looking into that first

donmccurdy:
@robertlong Would it be ok for me to merge this? Or do you think the bug in Third Room is related to this change?

robertlong:
The bug was unrelated. This should be safe to merge 👍

donmccurdy:
Ok, thank you!

PR #684, opened 2022-09-16
Fix attribute normalization in sortPrimitiveWeights()

### LINKED ISSUE #673: quantize: Broken output for skinned mesh
For this input...

[outfit-f-casual-01-v2.glb.zip](https://github.com/donmccurdy/glTF-Transform/files/9502437/outfit-f-casual-01-v2.glb.zip)

... quantization breaks the output:

```
#	CODE	MESSAGE	SEVERITY	POINTER
0	EMPTY_ENTITY	Entity cannot be empty.	0	/skins/0/joints
1	EMPTY_ENTITY	Entity cannot be empty.	0	/skins/1/joints
2	EMPTY_ENTITY	Entity cannot be empty.	0	/skins/2/joints
3	EMPTY_ENTITY	Entity cannot be empty.	0	/skins/3/joints
```

donmccurdy:
- Fixes #673

### DIFF
--- a/packages/functions/src/sort-primitive-weights.ts
+++ b/packages/functions/src/sort-primitive-weights.ts
@@ -43,8 +43,8 @@ export function sortPrimitiveWeights(prim: Primitive | PrimitiveTarget, limit = 
 	const dstJoints = new Uint32Array(setCount * 4);
 
 	for (let i = 0; i < vertexCount; i++) {
-		getVertexElements(prim, i, 'WEIGHTS', srcWeights);
-		getVertexElements(prim, i, 'JOINTS', srcJoints);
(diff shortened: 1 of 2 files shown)


donmccurdy:
Remaining issues: the asset renders correctly after quantization. Some error is visible but likely within expectations for default quantization settings. However the validator complains that the weights do not (quite) sum to 1, so there is a remaining issue in precisely normalizing uint8 and uint16 joint weights.

donmccurdy:
> Some imprecision is visible but likely within expectations for default quantization settings...

This turned out to be caused by the use of >4 skinning weights, with three.js does not yet support. The output appears fine in engines that do support 8 weights, like Babylon.js. The only issues that needed to be fixed by this PR were validation-related.

Issue #11, opened 2018-12-19
Function to flatten node hierarchy, merge meshes


donmccurdy:
Plan —

- [x] (1) Add `transformPrimitive(prim, matrix)`, `transformMesh(mesh, matrix)`
    - #700
    - #657
- [x] (2) Add `clearNodeParent(node)`, `clearNodeTransform(node)`. Requires (1).
    - Detaches node from its parent, applies world transform to `node.mesh`
    - #784
- [x] (3) Add `document.transform(flatten())`. Requires (1) and (2).
    - https://github.com/donmccurdy/glTF-Transform/pull/790 
    - Deletes unneeded empties (with options to keep them)
    - Keeps armatures
    - ~~Clears parents and transforms of nodes where possible~~
- [ ] (4) Add `joinPrimitives(prim: Primitive[])`.
    - #707 
- [ ] (5) Add `document.transform(join())`. Requires (1), (2), (3), (4).
    - Assumes node hierarchy has already been flattened as desired, using (3).
    - Joins all compatible meshes that are siblings in the scene hierarchy.
    - `TODO`: Maximum vertex counts per draw? Spatial locality?
    - #658

***

Other considerations —

Considerations:

- Update IBMs if node is a joint?
- ~~`clearNodeTransform` must warn on attached camera or light~~
- Apply world transforms to TRS animations?
- Apply world transforms to volumetric properties?
- Detach shared vertex streams? Deal with detached accessors?

donmccurdy:
Part of prune package?

dbuck:
I potentially could use this functionality also, though not fully flatten, I just want to merge a few of the meshes together, and am wondering whether I should do it in the obj before we get to gltf, or just tack it on to the pipeline scripts here ;)

In terms of how to psuedocode approach it, am I on track here?

Create new node/mesh/accessors to receive the merged details
targetPositionArray = accessor.getArray .. etc for each type

list nodes, filter to the ones I want to merge
from those nodes, get the meshes
from those meshes, get the accessors  for position/normal/uv
from the accessors get the arrays
concat them with the targetArray
dispose :all-the-things:

How should it handle textured meshes? Attach newly merged node to a scene, profit?

dbuck:
For the generalized transform, an input filter to select which items to merge would be helpful, ex: `(mesh: Mesh) => boolean` if Mesh is the correct level to select from? That would allow a user to: merge only meshes where mesh.name contains 'boots'

[...]

PR #707, opened 2022-10-18
Add joinPrimitives()

donmccurdy:
Given a list of compatible Mesh Primitives, returns new Primitive containing their vertex data. Compatibility requires that all Primitives share the same Materials, draw mode, and vertex attribute types. Primitives using morph targets cannot currently be joined.

Related:

- #11 

To do:

- [x] Implementation
- [x] Unit tests

Example:

```javascript
import { joinPrimitives } from '@gltf-transform/functions';

// Succeeds if Primitives are compatible, or throws an error.
const result = joinPrimitives(mesh.listPrimitives());

for (const prim of mesh.listPrimitives()) {
    prim.dispose();
}

mesh.addPrimitive(result);
```

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -45,6 +45,7 @@ export {
 export {
 	TypedArray,
 	TypedArrayConstructor,
+	ComponentTypeToTypedArray,
 	PropertyType,
 	Format,
 	Nullable,
--- a/packages/functions/src/index.ts
+++ b/packages/functions/src/index.ts
@@ -61,6 +61,7 @@ export * from './flatten';
 export * from './get-node-scene';
 export * from './inspect';
 export * from './instance';
+export * from './join-primitives';
 export * from './meshopt';
 export * from './metal-rough';
 export * from './normals';
--- a/packages/functions/src/join-primitives.ts
+++ b/packages/functions/src/join-primitives.ts
@@ -1,5 +1,13 @@
-import { Accessor, Document, Primitive, ComponentTypeToTypedArray } from '@gltf-transform/core';
-import { createIndices, createPrimGroupKey, isUsed } from './utils';
+import { Document, Primitive, ComponentTypeToTypedArray } from '@gltf-transform/core';
+import { createIndices, createPrimGroupKey } from './utils';
+
+interface JoinPrimitiveOptions {
+	skipValidation?: boolean;
+}
+
+const JOIN_PRIMITIVE_DEFAULTS: Required<JoinPrimitiveOptions> = {
+	skipValidation: false,
+};
 
 /**
  * Given a list of compatible Mesh {@link Primitive Primitives}, returns new Primitive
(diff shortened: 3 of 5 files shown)


drcmda:
in case it helps, these are the pruning steps in gltfjsx: https://github.com/pmndrs/gltfjsx/blob/master/src/utils/parser.js#L239-L310 pruning the graph results in smaller graphs but also improves performance since there are fewer nodes to parse and fewer matrix updates. 

```jsx
      /** Empty or no-property groups
       *    <group>
       *      <mesh geometry={nodes.foo} material={materials.bar} />
       *  Solution:
       *    <mesh geometry={nodes.foo} material={materials.bar} />
       */

      /** Double negative transforms
       *    <group rotation={[-Math.PI / 2, 0, 0]}>
       *      <group rotation={[Math.PI / 2, 0, 0]}>
       *        <mesh geometry={nodes.foo} material={materials.bar} />
       *  Solution:
       *    <mesh geometry={nodes.foo} material={materials.bar} />
       */
       
      /** Transform overlap      
       *    <group position={[10, 0, 0]} scale={2} rotation={[-Math.PI / 2, 0, 0]}>
       *      <mesh geometry={nodes.foo} material={materials.bar} />
       *  Solution:
       *    <mesh geometry={nodes.foo} material={materials.bar} position={[10, 0, 0]} scale={2} rotation={[-Math.PI / 2, 0, 0]} />
       */       
       
      /** Lack of content
       *    <group position={[10, 0, 0]} scale={2} rotation={[-Math.PI / 2, 0, 0]}>
       *      <group position={[10, 0, 0]} scale={2} rotation={[-Math.PI / 2, 0, 0]}>
       *        <group position={[10, 0, 0]} scale={2} rotation={[-Math.PI / 2, 0, 0]} />
       * Solution:
       *   (delete the sub graph)
       */
```

currently it is only a one-pass solution which is not enough. given:

```jsx
export function Model(props) {
  const { nodes, materials } = useGLTF('/scene_test-transformed.glb')
  return (
    <group {...props} dispose={null}>
      <group>
        <group rotation={[-Math.PI / 2, 0, 0]}>
          <group>
            <group rotation={[Math.PI / 2, 0, 0]}>
              <group position={[0.16, 0.79, -1.97]} rotation={[-0.54, 0.93, -1.12]} scale={0.5}>
                <mesh geometry={nodes.Object_4.geometry} material={materials.Texture} />
              </group>
            </group>
          </group>
        </group>
      </group>
    </group>
  )
}
```

it will currently prune it as:

```jsx
export function Model(props) {
  const { nodes, materials } = useGLTF('/scene_test-transformed.glb')
  return (
    <group {...props} dispose={null}>
      <group rotation={[-Math.PI / 2, 0, 0]}>
        <group rotation={[Math.PI / 2, 0, 0]}>
          <mesh geometry={nodes.Object_4.geometry} material={materials.Texture} position={[0.16, 0.79, -1.97]} rotation={[-0.54, 0.93, -1.12]} scale={0.5} />
        </group>
      </group>
    </group>
  )
}
```

which would be smaller if it had a backup pass:

```jsx
export function Model(props) {
  const { nodes, materials } = useGLTF('/scene_test-transformed.glb')
  return (
    <group {...props} dispose={null}>
      <mesh geometry={nodes.Object_4.geometry} material={materials.Texture} position={[0.16, 0.79, -1.97]} rotation={[-0.54, 0.93, -1.12]} scale={0.5} />
    </group>
  )
}
```

i think graph pruning would make more sense as a gltf-transform step and i would love to remove it.

drcmda:
[scene_test.glb.zip](https://github.com/donmccurdy/glTF-Transform/files/10279601/scene_test.glb.zip)

this a model for testing (the one above)

donmccurdy:
@drcmda — the depth of the scene graph can be reduced with the new `flatten()` transform in the v3 prerelease:

```javascript
import { flatten } from '@gltf-transform/functions';

await document.transform(flatten());
```

| before | after  |
|---|---|
| Screenshot 2023-01-25 at 8 50 49 PM | Screenshot 2023-01-25 at 8 50 54 PM |

This PR is still open though, as I'm trying to go a bit further and support merging mesh primitives to reduce the draw calls after the scene graph has been flattened.

drcmda:
@donmccurdy wow, nice work! if you have a minute to spare, could you go a little into how it's doing it and to which extend?

donmccurdy:
@drcmda sure thing, here's the description for `flatten()` —

> Flattens the scene graph, leaving Nodes with Meshes, Cameras, and other
attachments as direct children of the Scene. Skeletons and their
descendants are left in their original Node structure.
> 
> Animation targeting a Node or its parents will prevent that Node from being
moved.

- https://github.com/donmccurdy/glTF-Transform/pull/790

It computes the world matrix for each meaningful node, then moves it to the scene root without moving the node around in world space.

I'm open to leaving some escape hatches to preserve more structure where needed, perhaps based on node names, but for now this is aiming for the minimum possible nesting without breaking animation, which will be helpful once merging meshes is supported.

Should land in alpha 05, yes. 👍🏻

Issue #532, opened 2022-03-25
Documentation missing links to cross-package sources


donmccurdy:
**Describe the bug**

API documentation for `@gltf-transform/extensions` is missing source links to implementations coming from the `@gltf-transform/core` package.

**To Reproduce**

1. Visit https://gltf-transform.donmccurdy.com/classes/extensions.materialsvariants.html#isRequired
2. Scroll to `dispose` or `isRequired` definitions
3. Note they have broken source links like 

    Defined in glTF-Transform/packages/core/dist/extension.d.ts:67

**Expected behavior**

Source links should be correct and clickable.

**Additional context**

- #342

donmccurdy:
- Fixed by https://github.com/donmccurdy/glTF-Transform/pull/940.

Issue #461, opened 2021-12-27
Multi-language support (I18N)


donmccurdy:
I'd like to support translation / internationalization of the project documentation (https://gltf-transform.donmccurdy.com/). Doing so would require:

- [ ] Updates to the documentation generator
- [ ] Translation work (requires either volunteers or funding)

I can help with the documentation generator, but not so much with the translation. If others would like to get involved, please reach out.

Issue #342, opened 2021-08-26
Remove TypeDoc dependency


donmccurdy:
Currently documentation is generated with TypeDoc. I'd like to switch to [API Extractor](https://api-extractor.com/pages/overview/demo_docs/) for a few reasons:

- well-defined declarative JSON output
- easier to write/maintain a custom theme (see https://github.com/donmccurdy/glTF-Transform/pull/234)
- able to customize navigation without hand-writing each link
- hopefully more stable, have been hitting issues with TypeDoc historically, and had a major breaking change in patch release v0.21.3 with no fix planned

I think the idea here would be to:

1. Confirm I can get API Extractor JSON output for this monorepo.
2. Find or build an HTML/MD documentation generator (separate package)
3. Rewrite my theme for the generator
4. Put it all together

https://api-extractor.com/pages/setup/custom_docs/

***

Could also consider sticking with TypeDoc but switching to its JSON output mode. Most of the issues above are really complaints about the lack of a robust theming API.

- [How do you generate your TypeScript docs? • Reddit](https://old.reddit.com/r/typescript/comments/a3d46n/how_do_you_generate_your_typescript_docs/eb58rl6/)
- [Create Beautiful Documentation](https://blog.ironcorelabs.com/create-beautiful-documentation-7c66a2c5fa19)
- [Generating Documentation for TypeScript Projects](https://blog.cloudflare.com/generating-documentation-for-typescript-projects/)
- https://github.com/asakusuma/json-typescript-docs

donmccurdy:
After some experiments with both [API Extractor](https://api-extractor.com/pages/overview/demo_docs/) and [TS Morph](https://ts-morph.com/), I think TS Morph might be the more practical way to go here.

PR #734, opened 2022-11-23
Set up API Extractor

donmccurdy:
Sets up [API Extractor](https://api-extractor.com/), generating a JSON representation of the API model for use in other tools.

Related:

- #342
- #461
- #532


donmccurdy:
- https://github.com/donmccurdy/glTF-Transform/issues/342#issuecomment-1538442371

Issue #783, opened 2023-01-11
Properties should have (internal?) access to their Document or Graph


donmccurdy:
Some functions signatures (see [listTextureChannels](https://gltf-transform.donmccurdy.com/functions.html#listtexturechannels)) need to be passed a copy of the Document, for reasons that probably aren't obvious to the caller. You'd expect to be only to pass in a Texture and nothing more.

Properties don't expose their Graph, though:

https://github.com/donmccurdy/glTF-Transform/blob/5e856765eaba5d298e18e2045196128c4cbfb3cd/packages/core/src/properties/property.ts#L55-L60

And the Document is the only way to access that from a function body. We frequently need it to trace dependencies in the dependency graph. There should be some way to access the Graph, given a Property. Optionally, access might go through the parent Document.

This would allow us to remove Documents from function parameters in several cases.

donmccurdy:
Fixed by https://github.com/donmccurdy/glTF-Transform/pull/787.

PR #787, opened 2023-01-19
Avoid requiring document as a function parameter unnecessarily

donmccurdy:
Removes document from the following method signatures:

- listTextureChannels(~~document,~~ texture)
- getTextureChannelMask(~~document,~~ texture)
- listTextureInfo(~~document,~~ texture)
- listTextureSlots(~~document,~~ texture)


donmccurdy:
Fixes https://github.com/donmccurdy/glTF-Transform/issues/783.

PR #790, opened 2023-01-19
Add flatten()

donmccurdy:
Progress:

- https://github.com/donmccurdy/glTF-Transform/issues/11

Remaining:

- [x] Unit tests

Design notes:

- Nodes containing Skins, Meshes, Cameras, Lights, etc. are hoisted to become Scene children
- Ineligible Nodes include:
    - Nodes affected by TRS animation (self or inherited)
    - Children of joints (e.g. joint hierarchies, mesh attachments to joints)
- Node transforms are updated in place, not baked into mesh vertices
- Certain edge cases are avoided here, but may become relevant with `join()` later
    - Apply transforms to volumetric materials
    - Dealing with mixed vertex stream structures


donmccurdy:
Aside - A `keepNames` parameter should just ensure that Nodes and Meshes with names are not consolidated into a new Node with a single Mesh and many Primitives. They'll still end up direct children of the Scene. I'd overthought that originally, in trying to preserve nested relationships between named Nodes.

Issue #747, opened 2022-12-18
More precise names for Texture, ImageUtils methods


donmccurdy:
The [Texture](https://gltf-transform.donmccurdy.com/classes/core.texture.html) and [ImageUtils](https://gltf-transform.donmccurdy.com/classes/core.imageutils.html) classes use "size" in method names ambiguously:

- `texture.getSize()`
- `ImageUtils.getSize()`
- `ImageUtils.getMemSize()`

None of these names are great, IMHO. Consider using a `*ByteLength` suffix when referring to quantities of bytes, or `getResolution` or `getDimensions` when referring to quantities of pixels.

Because this would be a breaking change, it would need to be done in the v3 release.

PR #797, opened 2023-01-24
ImageUtils: Rename getMemSize → getGPUByteLength

donmccurdy:
- Fixes https://github.com/donmccurdy/glTF-Transform/issues/747

Small breaking change for users who:

- (a) use ImageUtils directly
- (b) implement an extension providing support for a non-default image format

I think both of these cases are pretty rare, so this should be a modest enough solution to the naming issue described in #747. It'd be more disruptive to rename getSize to something like getDimensions, I don't feel like that change is really worth it at this point.

### DIFF
--- a/packages/core/src/utils/image-utils.ts
+++ b/packages/core/src/utils/image-utils.ts
@@ -131,7 +131,7 @@ export class ImageUtils {
 	}
 
 	/** Returns a conservative estimate of the GPU memory required by this image. */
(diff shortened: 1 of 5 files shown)


donmccurdy:
Renamed getGPUByteLength to getVRAMByteLength in https://github.com/donmccurdy/glTF-Transform/pull/812. This PR wasn't released with v3.0 yet, so not a breaking change to rename it again.

PR #803, opened 2023-01-29
Improve weld() performance with spatial grid

### LINKED ISSUE #738: Improve performance of weld() function
**Describe the bug**
document transform "weld" function never finishes, or takes a very, very long time.

**To Reproduce**
1. Load attached GLB: `const doc = await io.read( 'weld-problem.glb' );`
2. `await doc.transform( weld() )`;

**Expected behavior**
The weld should finish after a reasonable time.

**Versions:**
 - 2.4.4
 - Win 10, node 16.17,0

**Additional context**
[weld-problem.glb.zip](https://github.com/donmccurdy/glTF-Transform/files/10134739/weld-problem.zip)

### LINKED ISSUE #870: weld() causes some meshes to have missing faces
I've noticed some missing faces when using the weld function. The tolerance has been set really low to minimise errors but doesn't seem to help. There is an attachment zip with both models. 

With weld + dedup
image

Without weld + dedup
image

[models.zip](https://github.com/donmccurdy/glTF-Transform/files/11042198/models.zip)

donmccurdy:
Hashes vertices against a spatial grid, making the search for weld candidates much faster. In 'exhaustive' mode we search all neighboring cells in the grid, otherwise we weld only against vertices in the same grid. Also fixes bugs in the earlier implementation, and ensures that we remove any degenerate geometry created by the welding process.

For current test case, "Lovecraftian" with 800,000 vertices with ~75% weld rate, this improves weld performance from 22s initially to 5s (`exhaustive: true`) or 1.5s (`exhaustive: false`).

- Fixes #738
- Fixes #870

### DIFF
--- a/packages/functions/src/weld.ts
+++ b/packages/functions/src/weld.ts
@@ -8,10 +8,10 @@ import {
 	TypedArray,
 	vec3,
 } from '@gltf-transform/core';
-import { cleanPrimitive } from './clean-primitive';
-import { dedup } from './dedup';
-import { prune } from './prune';
-import { createIndices, createTransform, formatDeltaOp } from './utils';
+import { cleanPrimitive } from './clean-primitive.js';
+import { dedup } from './dedup.js';
+import { prune } from './prune.js';
+import { createIndices, createTransform, formatDeltaOp } from './utils.js';
 
 // DEVELOPER NOTES: Ideally a weld() implementation should be fast, robust,
 // and tunable. The writeup below tracks my attempts to solve for these
--- a/packages/functions/src/weld.ts
+++ b/packages/functions/src/weld.ts
@@ -5,12 +5,42 @@ import {
 	PrimitiveTarget,
 	PropertyType,
 	Transform,
-	TransformContext,
 	TypedArray,
 	vec3,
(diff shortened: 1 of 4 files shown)


packages/functions/src/weld.ts
```diff
@@ -239,6 +277,8 @@ function _weldPrimitive(doc: Document, prim: Primitive, options: Required<WeldOp
 			swapAttributes(target, srcAttr, writeMap, dstVertexCount);
 		}
 	}
+
+	console.timeEnd('weldPrimitive');
```
donmccurdy, line 281: TODO: remove debugging output

donmccurdy:
- weld, before: 228,297 vertices
- welf, after: 228,498 vertices

donmccurdy:
Adding test models from #870: https://github.com/donmccurdy/glTF-Transform/files/11042198/models.zip

Note that for this model, the updates in this PR improve the output (vertices that shouldn't have been welded are no longer welded), but some need for cleanup remains:

- [ ] Remove degenerate triangles (collapsed to ≤2 vertices)
- [ ] Remove degenerate primitives and meshes (collapsed to 0 GL primitives)

PR #833, opened 2023-02-24
Allow Nodes in multiple Scenes, deprecate Node#getParent()

### LINKED ISSUE #831: Allow root nodes in multiple scenes
Not sure if this is a feature request or a bug report.

The glTF spec says that "The same root node can appear in multiple scenes"  and this seems to work in most tools I've loaded my glTFs in.   But your tool always validates scenes and removes any nodes that appear in multiple scenes. 

Screen Shot 2023-02-23 at 3 25 45 PM

donmccurdy:
Adds support for reusing the same Node as a root Node in multiple Scenes. Deprecates inappropriate methods based on previous assumptions of single-Scene nodes.

**Changes:**

- Fixes #831 
- Deprecate `Node.getParent() : Scene | Node | null`
- Deprecate `getNodeScene(node) : Scene | null`
- Add `Node.getParentNode() : Node | null`
- Add `listNodeScenes(node) : Scene[]`


donmccurdy:
This PR introduced a regression (#1025), to be fixed in #1026.

Issue #837, opened 2023-02-27
Hope to add options to control "flatten" and "join" for "optimize" command


xlsfs:
Want to preserve hierarchy when optimizing

PR #838, opened 2023-02-27
Update cli.ts

xlsfs:
#837 
add options to control "flatten" and "join" for "optimize" command


donmccurdy:
Thanks! I think this sounds good but I'm AFK for a bit, and will need to take a look this weekend.

donmccurdy:
Thank you!

PR #840, opened 2023-03-03
update readme forgotten imports

y-71:
Some imports were missing on the `readme`

not a big deal but it's also cool when the code runs when you copy it


donmccurdy:
Thank you!

PR #868, opened 2023-03-22
Draft: Add vertexPaint()

### LINKED ISSUE #802: Add vertexPaint()
In scenes with many objects like those found in CAD or low-poly scenes, we may have 100s of materials with solid materials. It can be advantageous join them, with the new join() function, but to do that we need to unify the materials. A new function that converts solid colors to vertex colors, and then de-duplicates materials, could be a big help here.

Example:

```javascript
import { vertexPaint, ... } from '@gltf-transform/functions';

await document.transform(
  vertexPaint(),
  dedup(),
  flatten(),
  join(),
  weld(),
  simplify({ ... }),
  draco()
);
```

donmccurdy:
- fixes #802 

Remaining:

- [ ] Complete implementation and TODOs
- [ ] Manual testing
- [ ] Decide on f32 vs u16 output
- [ ] Bilinear interpolation (https://github.com/scijs/ndarray-linear-interpolate)
- [ ] Unit tests

Test cases:

- https://sketchfab.com/3d-models/low-poly-environment-2e18c1baa9164093ad2e99e0a904363a

### DIFF
--- a/packages/functions/src/vertex-paint.ts
+++ b/packages/functions/src/vertex-paint.ts
@@ -1,0 +1,154 @@
+import {
+	ColorUtils,
+	Document,
+	Material,
+	Primitive,
+	PropertyType,
+	Texture,
+	TextureInfo,
+	Transform,
+	vec2,
+	vec3,
+} from '@gltf-transform/core';
+import { getPixels } from 'ndarray-pixels';
+import { createTransform } from './utils.js';
+import type { NdArray } from 'ndarray';
+import { prune } from './prune.js';
+
+const NAME = 'vertexPaint';
+
+/** Options for the {@link colorspace} function. */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface VertexPaintOptions {}
+
+const VERTEX_PAINT_DEFAULTS: VertexPaintOptions = {};
+
+/**
+ * Paints each {@link Material}'s base color texture to vertex colors, and removes the texture.
+ *
+ * Example:
+ *
+ * ```javascript
+ * import { vertexPaint } from '@gltf-transform/functions';
+ *
+ * const mesh = document.getRoot().listMeshes()
+ * 	.find((mesh) => mesh.getName() === 'MyMesh');
+ * const prim = mesh.listPrimitives()[0];
+ *
+ * prim.getAttribute('COLOR_0'); // → null
+ *
+ * await document.transform(vertexPaint());
+ *
+ * prim.getAttribute('COLOR_0'); // → Accessor
+ * ```
+ */
+export function vertexPaint(_options: VertexPaintOptions = VERTEX_PAINT_DEFAULTS): Transform {
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	const options = { ...VERTEX_PAINT_DEFAULTS, ..._options } as Required<VertexPaintOptions>;
+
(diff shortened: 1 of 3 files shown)


packages/functions/src/vertex-paint.ts
```diff
@@ -0,0 +1,152 @@
...
+function _vertexPaintPrimitive(
+	document: Document,
+	prim: Primitive,
+	pixels: NdArray<Uint8Array>,
+	texCoord: number
+): void {
+	const uv = prim.getAttribute(`TEXCOORD_${texCoord}`)!;
+	const color = document
+		.createAccessor()
+		.setArray(new Float32Array(uv.getCount() * 3))
```
kzhsw, line 95: Since `Uint8Array` is used here, maybe it would be better to use quantized color attributes, or give user an option to do that.

packages/functions/src/vertex-paint.ts
```diff
@@ -0,0 +1,152 @@
...
+		.setType('VEC3'); // TODO(bug): Consider opacity?
+
+	const _uv = [0, 0] as vec2;
+	const _color = [0, 0, 0] as vec3;
+
+	// Read color from texture, write to COLOR_0 vertex attribute.
+	for (let i = 0, il = uv.getCount(); i < il; i++) {
+		uv.getElement(i, _uv);
+		const x = Math.floor(_uv[0] * pixels.shape[0]);
+		const y = Math.floor(_uv[1] * pixels.shape[1]);
```
kzhsw, line 106: Could `bilinear` and `bicubic` be optional used here?

packages/functions/src/vertex-paint.ts
donmccurdy: Will likely do uint16 or float32 as the default, and then the `quantize()` function can be used to quantize afterward if desired. Note that vertex colors are linear and the texture was sRGB, so choosing uint8 vertex colors will lose precision compared to the uint8 texture values.

packages/functions/src/vertex-paint.ts
donmccurdy: Possibly — I think adding bilinear interpolation would be easy enough at least. For use cases I have in mind, the interpolation probably won't matter very much. If you think this affects your models, test cases would be welcome!

kzhsw:
Also consider baking normal texture into vertex normal with optional quantization.

packages/functions/src/vertex-paint.ts
```diff
@@ -0,0 +1,152 @@
...
+
+		// Pre-process; load pixel data into memory.
+		for (const material of root.listMaterials()) {
+			const texture = material.getBaseColorTexture();
+			const textureInfo = material.getBaseColorTextureInfo();
+			if (!texture || !textureInfo) continue;
+
+			const pixels = (await getPixels(texture.getImage()!, texture.getMimeType())) as NdArray<Uint8Array>;
+			const texCoord = textureInfo.getTexCoord();
+			pixelsLookup.set(material, { pixels, texCoord });
```
kzhsw, line 63: Would [KHR_texture_transform](https://github.com/donmccurdy/glTF-Transform/blob/main/packages/extensions/src/khr-texture-transform/texture-transform.ts) being supported here?

kzhsw:
Maybe this can also paint normal texture into vertex normals

donmccurdy:
Closing for now – I think this needs (in addition to time...) a clearer motivating use case to make decisions about the implementation and to test it properly.

Issue #501, opened 2022-02-16
Broken animation after resampling


donmccurdy:
**Describe the bug**

After resampling, the animation sequence in this example is broken.

**To Reproduce**

[Archive.zip](https://github.com/donmccurdy/glTF-Transform/files/8081046/Archive.zip)

1. `gltf-transform resample Arm.glb Arm-resampled.glb`

Decreasing tolerance will fix the issue, but also prevents resampling from reducing the size very much. My hunch is that the [current implementation of `resample()](https://github.com/donmccurdy/glTF-Transform/blob/ca18796/packages/functions/src/resample.ts#L72-L142) does not handle quaternion keyframe tracks properly.

**Expected behavior**

Animation should appear visually identical, within some small tolerance.

**Related**

- https://github.com/donmccurdy/glTF-Transform/pull/492
- https://github.com/donmccurdy/glTF-Transform/issues/489

**Sub-issues:**

- [ ] `resample()` should remove channels that have no effect
- [x] `resample()` must slerp, not lerp, quaternion tracks
- [x] `resample()` must not collapse keyframes A-B-C if A and C are >180º apart

donmccurdy:
Digging into the file a bit more, there are lots of junk keyframe tracks that aren't contributing anything to the final animation. Some are small (2 keyframes,  both identical) and others are large (50,627 keyframes, all containing `scale=1,1,1`). The `resample` command deals with the duplicate keyframes but doesn't currently prune no-op tracks. That — along with an automatic `dedup` step — would probably be reasonable additions to the function.

<img width="469" alt="Screen Shot 2022-02-16 at 11 03 06 AM" src="https://user-images.githubusercontent.com/1848368/154305910-f63a364a-7015-435f-b4f2-c54a618b3378.png">

hybridherbst:
To chime in here regarding the file - everything with two keyframes is there because the "animation pose" is different to the "non-animated pose" for this model. The rest pose could be anything (e.g. stretched out) but if during the entire animation clip a part of the model would be rotated differently, then there's a start and end keyframe for that but no actual animation inbetween. So, these tracks aren't always "no-op tracks", they can also be "pose tracks" without any animation/movement happening.

donmccurdy:
Somehow gltfpack manages to reduce this file to just 6 animation channels – I guess that must (at least in part) mean it's baking the animated pose to the rest pose? Does gltfpack's output work for your use cases, or is that something you'd want to be opt-in if it happens at all?

[...]

Issue #908, opened 2023-04-14
core: optimize the performance of isRefMap


kzhsw:
**Is your feature request related to a problem? Please describe.**
In a recent benchmark on resample function of gltf-transfrom git ([618c506](https://github.com/donmccurdy/glTF-Transform/commit/618c506920dd49c49ec36287febac934b3b87d95)), I found that the function `isRefMap` consumes most (26.8%) of CPU time:
![chrome performance profiler](https://user-images.githubusercontent.com/120431159/231983312-c0f80e7c-3dca-460a-abe5-0ceb3afec0ef.png)

**Describe the solution you'd like**
Replace isRefMap to code below, which runs much faster according to benchmark:
```typescript
const hasOwnProperty = Object.prototype.hasOwnProperty;

export function isRefMap(value: unknown): boolean {
	if (!isPlainObject(value)) {
		return false;
	}
	const objectValue = value as object;
	for (const k in objectValue) {
		if (!hasOwnProperty.call(value, k)) {
			continue;
		}
		return objectValue[k as keyof typeof objectValue] as object instanceof GraphEdge;
	}
	return false;
}
```
This could be done by changing `packages/core/src/utils/ref-utils.ts` file, and replace import of `isRefMap` in file `packages/core/src/properties/property.ts`.
![chrome performance profiler after patch](https://user-images.githubusercontent.com/120431159/231986286-4ce26392-ed7a-45bc-bb85-a7c7b3c56242.png)

**Describe alternatives you've considered**
* Impl this in [property-graph](https://github.com/donmccurdy/property-graph) repo instead of changing glTF-Transform.
* Monkey patch `property-graph` at runtime.
* Submit this to v8 or spidermonkey issue tracker and hope they would optimize this `Object.values(value)[0]`.

**Additional context**
The model used for benchmark: Arm.glb from [Archive.zip](https://github.com/donmccurdy/glTF-Transform/files/8081046/Archive.zip) from https://github.com/donmccurdy/glTF-Transform/issues/501

Also, according to the benchmark result above, performance of `GraphNode#get()` might need futher optimizations.

The benchmark code, which runs in the root dir of a built glTF-Transform repo :
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
</head>
<body>

<input type="file" id="file0">
<button type="button" id="btn0">Click</button>
<script type="importmap">
{
  "imports": {
[...]

PR #909, opened 2023-04-15
resample: only compute slerp when necessary

kzhsw:
Currently when resampling quaternion keyframes, both `angle` and `sample` are computed, but not all of them are used for `keep`, for example, if result of `angle` is larger `Math.PI - Number.EPSILON`, value of `sample` is not needed since `keep` would be always true.

Alternative:
Only compute `angle` when necessary, compute `sample` using slerp first, and compute `angle` if they are close enough.

See also <https://github.com/donmccurdy/glTF-Transform/issues/501>

### DIFF
--- a/packages/functions/src/resample.ts
+++ b/packages/functions/src/resample.ts
@@ -109,9 +109,13 @@ function optimize(sampler: AnimationSampler, path: GLTF.AnimationChannelTargetPa
 
 			if (interpolation === 'LINEAR' && path === 'rotation') {
 				// Prune keyframes colinear with prev/next keyframes.
-				const sample = slerp(tmp as quat, valuePrev as quat, valueNext as quat, t) as number[];
 				const angle = getAngle(valuePrev as quat, value as quat) + getAngle(value as quat, valueNext as quat);
-				keep = !MathUtils.eq(value, sample, tolerance) || angle + Number.EPSILON >= Math.PI;
+				if (angle + Number.EPSILON >= Math.PI) {
+					keep = true;
+				} else {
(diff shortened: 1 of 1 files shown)


donmccurdy:
Thank you @kzhsw! Please see feedback in https://github.com/donmccurdy/glTF-Transform/issues/908#issuecomment-1510475423.

kzhsw:
before
after

| function | before (ms) | after (ms) | diff (ms) |
| :---: | ---: | ---: | ---: |
| on (getAngle) |144.9 | 179.1 |  +34.2|
| rn (slerp)| 264.4 | 196.4  | -68.0|
| total| 409.3 | 375.5 | -33.8|

Benchmarked using the html code in #908.

donmccurdy:
@kzhsw Sorry to hold this up — When I test the `Arm.glb` file mentioned in the issue, with `console.time` blocks around resample(), total time is essentially the same before and after this change. If I check "total blocking time" at the bottom of your screenshots above, it appears similar before and after as well. I see the functions you mention are consuming less time, but I'm not sure the `resample()` function is faster as a result.

Would it be possible to share a glTF file that benefits from the change, in case I need to refactor this code later? Or have the changes in #910 done what was required here?

kzhsw:
Yes, that is a minor optimization comparing to #910, after which [GraphNode#get()](https://github.com/donmccurdy/property-graph/blob/935f4b6b89cb7d232e3d6a5a4374fb5ab573962a/src/graph-node.ts#L178) is taking most of the CPU time, feel free to close this.

Here is the screenshot of 20 runs in chromium, using console.time:
before
before
after
after

donmccurdy:
Ok, I see - thank you! I'll close this out for now, but remain open to ways of improving performance here.

PR #934, opened 2023-05-02
resample: Port keyframe resampling to WebAssembly

### LINKED ISSUE #290: Resampling should include morph target animations
**Describe the bug**
Seems that resampling is missing some condition for removing/keeping keyframes.
The attached animation breaks after resampling.

**To Reproduce**
1. Download 
[pfc-head-resampling-issue.zip](https://github.com/donmccurdy/glTF-Transform/files/6816761/pfc-head-resampling-issue.zip)
2. Run `gltf-transform resample pfc-head-resampling-issue.glb out.glb`
3. Watch both the original and the result in https://sandbox.babylonjs.com/

**Expected behavior**
Resampled animation looks the same as the original animation but might be smaller file size.
Actual: resampled animation is smaller but looks very different / wrong; mesh breaks in some places.

**Additional context**
Original vs. Resampled

### LINKED ISSUE #922: resample: optimizing performance of GraphNode#get()
After profiling `resample` with chrome's devtools on git master (89108e04), it could be found that `GraphNode#get()` is taking most (33.5%) of the CPU time.
chrome devtools profiling screenshot
time of last 10 runs

Replace the calls to `Accessor#getScalar`, `Accessor#setScalar`, `Accessor#getElement`, `Accessor#setElement` to more direct access to the underlying array.

<details><summary>The patch</summary>
<p>

```patch
diff --git a/packages/functions/src/resample.ts b/packages/functions/src/resample.ts
index fca4f93..9d3dcd5 100644
--- a/packages/functions/src/resample.ts
+++ b/packages/functions/src/resample.ts
@@ -8,6 +8,7 @@ import {
        Root,
        Transform,
        TransformContext,
+       TypedArray,
 } from '@gltf-transform/core';
 import quat, { getAngle, slerp } from 'gl-matrix/quat';
 import { dedup } from './dedup.js';
@@ -90,22 +91,25 @@ function optimize(sampler: AnimationSampler, path: GLTF.AnimationChannelTargetPa
        const value: number[] = [];
        const valueNext: number[] = [];
        const valuePrev: number[] = [];
+       const inputArray = input.getArray()!;
+       const outputArray = output.getArray()!;
+       const outputElementSize = output.getElementSize();

        let writeIndex = 1;

        for (let i = 1; i < lastIndex; ++i) {
-               const timePrev = input.getScalar(writeIndex - 1);
-               const time = input.getScalar(i);
-               const timeNext = input.getScalar(i + 1);
+               const timePrev = inputArray[writeIndex - 1];
+               const time = inputArray[i];
+               const timeNext = inputArray[i + 1];
                const t = (time - timePrev) / (timeNext - timePrev);

                let keep = false;

                // Remove unnecessary adjacent keyframes.
                if (time !== timeNext && (i !== 1 || time !== input.getScalar(0))) {
-                       output.getElement(writeIndex - 1, valuePrev);
-                       output.getElement(i, value);
-                       output.getElement(i + 1, valueNext);
+                       getElement(outputElementSize, outputArray, writeIndex - 1, valuePrev);
+                       getElement(outputElementSize, outputArray, i, value);
+                       getElement(outputElementSize, outputArray, i + 1, valueNext);

                        if (interpolation === 'LINEAR' && path === 'rotation') {
                                // Prune keyframes colinear with prev/next keyframes.
@@ -125,8 +129,9 @@ function optimize(sampler: AnimationSampler, path: GLTF.AnimationChannelTargetPa
                // In-place compaction.
                if (keep) {
                        if (i !== writeIndex) {
-                               input.setScalar(writeIndex, input.getScalar(i));
-                               output.setElement(writeIndex, output.getElement(i, tmp));
+                               inputArray[writeIndex] = inputArray[i];
+                               setElement(outputElementSize, outputArray, writeIndex,
+                                       getElement(outputElementSize, outputArray, i, tmp));
                        }
                        writeIndex++;
                }
@@ -134,15 +139,16 @@ function optimize(sampler: AnimationSampler, path: GLTF.AnimationChannelTargetPa

        // Flush last keyframe (compaction looks ahead).
        if (lastIndex > 0) {
-               input.setScalar(writeIndex, input.getScalar(lastIndex));
-               output.setElement(writeIndex, output.getElement(lastIndex, tmp));
+               inputArray[writeIndex] = inputArray[lastIndex];
+               setElement(outputElementSize, outputArray, writeIndex,
+                       getElement(outputElementSize, outputArray, lastIndex, tmp));
                writeIndex++;
        }

        // If the sampler was optimized, truncate and save the results. If not, clean up.
        if (writeIndex !== input.getCount()) {
-               input.setArray(input.getArray()!.slice(0, writeIndex));
-               output.setArray(output.getArray()!.slice(0, writeIndex * output.getElementSize()));
+               input.setArray(inputArray.slice(0, writeIndex));
+               output.setArray(outputArray.slice(0, writeIndex * outputElementSize));
                sampler.setInput(input);
                sampler.setOutput(output);
        } else {
@@ -151,6 +157,19 @@ function optimize(sampler: AnimationSampler, path: GLTF.AnimationChannelTargetPa
        }
 }

+function getElement(elementSize: number, array: TypedArray, index: number, target: number[]): number[] {
+       for (let i = 0; i < elementSize; i++) {
+               target[i] = array[index * elementSize + i];
+       }
+       return target;
+}
+
+function setElement(elementSize: number, array: TypedArray, index: number, value: number[]): void {
+       for (let i = 0; i < elementSize; i++) {
+               array[index * elementSize + i] = value[i];
+       }
+}
+
 function lerp(v0: number, v1: number, t: number): number {
        return v0 * (1 - t) + v1 * t;
 }


```

</p>
</details> 

chrome devtools after patch
last 10 runs after patch

Reimplement the resample algorithm in wasm, using wasm and simd can make it ~2x faster than the optimized code for the same model.

**Additional context**
The model used for benchmark: Arm.glb from [Archive.zip](https://github.com/donmccurdy/glTF-Transform/files/8081046/Archive.zip) from https://github.com/donmccurdy/glTF-Transform/issues/501

The benchmark code is at https://github.com/donmccurdy/glTF-Transform/issues/908

This might benefit from futher optimizing such as inlining and loop unrolling, or more directly using TypedArray#set.

Support of normalized animation is missing here.

donmccurdy:
As discussed in #922, the resample() function currently spends a lot of time in resource graph operations. Rewriting it to operate directly on the underlying accessor arrays improves performance dramatically (170ms → 60ms for my 800,000 keyframe test dataset on M1 mac). Rewriting the logic in optimized WebAssembly gets even greater savings (170ms → 20ms).

The WebAssembly code is compiled from AssemblyScript, and will be maintained in a separate repository (https://github.com/donmccurdy/keyframe-resample-wasm).  

***

- Fixes #922 
- Fixes #290 

***

Major thanks to @kzhsw for extensive research, prototyping, and advice in #922.

### DIFF
--- a/packages/functions/src/resample.ts
+++ b/packages/functions/src/resample.ts
@@ -3,29 +3,49 @@ import {
 	AnimationSampler,
 	Document,
 	GLTF,
+	MathUtils,
 	PropertyType,
 	Root,
 	Transform,
 	TransformContext,
+	TypedArray,
 } from '@gltf-transform/core';
 import { dedup } from './dedup.js';
 import { createTransform, isTransformPending } from './utils.js';
-import { ready, resampleWASM } from 'keyframe-resample';
+import { resampleDebug } from 'keyframe-resample';
 
 const NAME = 'resample';
 
 export interface ResampleOptions {
+	ready?: Promise<void>;
+	resample?: typeof resampleDebug;
 	tolerance?: number;
 }
 
-const RESAMPLE_DEFAULTS: Required<ResampleOptions> = { tolerance: 1e-4 };
+const RESAMPLE_DEFAULTS: Required<ResampleOptions> = {
+	ready: Promise.resolve(),
+	resample: resampleDebug,
+	tolerance: 1e-4,
+};
 
 /**
  * Resample {@link Animation}s, losslessly deduplicating keyframes to reduce file size. Duplicate
  * keyframes are commonly present in animation 'baked' by the authoring software to apply IK
  * constraints or other software-specific features. Based on THREE.KeyframeTrack.optimize().
  *
- * Example: (0,0,0,0,1,1,1,0,0,0,0,0,0,0) --> (0,0,1,1,0,0)
+ * Result: (0,0,0,0,1,1,1,0,0,0,0,0,0,0) → (0,0,1,1,0,0)
+ *
+ * Example:
+ *
+ * ```
+ * import { ready, resample } from 'keyframe-resample';
+ *
+ * // JavaScript (slower)
+ * await document.transform(resample());
+ *
+ * // WebAssembly (faster)
+ * await document.transform(resample({ ready, resample }));
+ * ```
  */
 export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transform {
 	const options = { ...RESAMPLE_DEFAULTS, ..._options } as Required<ResampleOptions>;
@@ -34,6 +54,9 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
 		const accessorsVisited = new Set<Accessor>();
 		const srcAccessorCount = document.getRoot().listAccessors().length;
 		const logger = document.getLogger();
+
+		const ready = options.ready;
+		const resample = options.resample;
 
 		await ready;
 
@@ -53,19 +76,19 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
 					accessorsVisited.add(input);
 					accessorsVisited.add(output);
 
-					const times = input.getArray()!.slice() as Float32Array; // TODO(cleanup)
-					const values = output.getArray()!.slice() as Float32Array; // TODO(cleanup)
+					const times = toFloat32Array(input.getArray()!, input.getComponentType(), input.getNormalized());
+					const values = toFloat32Array(output.getArray()!, input.getComponentType(), input.getNormalized());
 					const elementSize = values.length / times.length;
 
 					const srcCount = times.length;
 					let dstCount: number;
 
 					if (samplerInterpolation === 'STEP') {
-						dstCount = resampleWASM(times, values, 'step', options.tolerance);
+						dstCount = resample(times, values, 'step', options.tolerance);
 					} else if (samplerTargetPaths.get(sampler) === 'rotation') {
-						dstCount = resampleWASM(times, values, 'slerp', options.tolerance);
+						dstCount = resample(times, values, 'slerp', options.tolerance);
 					} else {
-						dstCount = resampleWASM(times, values, 'lerp', options.tolerance);
+						dstCount = resample(times, values, 'lerp', options.tolerance);
 					}
 
 					if (dstCount < srcCount) {
@@ -90,4 +113,19 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
 
 		logger.debug(`${NAME}: Complete.`);
 	});
+}
+
+function toFloat32Array(
+	srcArray: TypedArray,
+	componentType: GLTF.AccessorComponentType,
+	normalized: boolean
+): Float32Array {
+	const dstArray = new Float32Array(srcArray);
+	if (!normalized) return dstArray;
+
+	for (let i = 0; i < dstArray.length; i++) {
+		dstArray[i] = MathUtils.decodeNormalizedInt(dstArray[i], componentType);
+	}
+
+	return dstArray;
 }
--- a/packages/functions/src/resample.ts
+++ b/packages/functions/src/resample.ts
@@ -3,15 +3,14 @@ import {
 	AnimationSampler,
 	Document,
 	GLTF,
-	MathUtils,
 	PropertyType,
 	Root,
 	Transform,
 	TransformContext,
 } from '@gltf-transform/core';
-import quat, { getAngle, slerp } from 'gl-matrix/quat';
 import { dedup } from './dedup.js';
 import { createTransform, isTransformPending } from './utils.js';
+import { ready, resampleWASM } from 'keyframe-resample';
 
 const NAME = 'resample';
 
@@ -36,6 +35,8 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
 		const srcAccessorCount = document.getRoot().listAccessors().length;
 		const logger = document.getLogger();
 
+		await ready;
+
 		let didSkipMorphTargets = false;
 
 		for (const animation of document.getRoot().listAnimations()) {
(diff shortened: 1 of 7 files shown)


packages/functions/src/resample.ts
```diff
@@ -46,14 +47,40 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
...
+
+				if (samplerInterpolation === 'STEP' || samplerInterpolation === 'LINEAR') {
+					const input = sampler.getInput()!;
+					const output = sampler.getOutput()!;
+
+					accessorsVisited.add(input);
+					accessorsVisited.add(output);
+
+					const times = input.getArray() as Float32Array;
+					const values = output.getArray() as Float32Array;
```
donmccurdy, line 66: TODO: If output array is not f32, we'll need to transfer it.

packages/functions/src/resample.ts
```diff
@@ -46,14 +47,40 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
...
+						dstCount = resampleWASM(times, values, 'step');
+					} else if (targetPath === 'rotation') {
+						dstCount = resampleWASM(times, values, 'slerp');
+					} else {
+						dstCount = resampleWASM(times, values, 'lerp');
+					}
+
+					if (dstCount < srcCount) {
+						sampler.setInput(input.clone().setArray(times.slice(0, dstCount)));
+						sampler.setOutput(output.clone().setArray(values.slice(0, dstCount * elementSize)));
```
donmccurdy, line 82: TODO: The original arrays have already been overwritten at this point. Decide if we want to preemptively clone the arrays in JS before resampling, or to change the WASM API.

packages/functions/src/resample.ts
```diff
@@ -46,14 +47,40 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
 			}
 
 			for (const sampler of animation.listSamplers()) {
-				if (samplerTargetPaths.get(sampler) === 'weights') {
+				const targetPath = samplerTargetPaths.get(sampler)!;
+				if (targetPath === 'weights') {
```
donmccurdy, line 51: TODO: It's unfortunate that morph target weights are still skipped here, and it'd be nice to figure out the cause and fix that. See #290.

kzhsw:
Things to consider:
* [keyframe-resample-wasm](https://github.com/donmccurdy/keyframe-resample-wasm) uses a [shared](https://github.com/donmccurdy/keyframe-resample-wasm/blob/d6896fa73af70370abf711486cdcc9c0b9b861ba/src/resample_bindings.ts#L32) wasm instance, maybe users can be given options to release it to release memory.
* Inline the wasm in js like [meshoptimizer](https://github.com/zeux/meshoptimizer) for old bundlers
* Multithreading for large chunks or different channels
* Fallback to js impl on unsupported plarform of user's choice

donmccurdy:
There are several other WASM dependencies in the project now (Draco, Meshopt, MikkTSpace), so this module isn't necessarily the limiting factor on bundler and platform support.

I think I'll need to draw a line somewhere on the complexity, or this isn't going to get finished.  😇 If you (or another reader) needs one of those things, I'd be happy to discuss tradeoffs around it. But unless there's a clear need, I'd like to keep this from getting any more complex than it is now.

packages/functions/src/resample.ts
kzhsw: In [spec](https://github.com/KhronosGroup/glTF/blob/d202b79a816943beacdd9be6c936b63a1a9c0ea5/specification/2.0/Specification.adoc?plain=1#L2555), animation samplers targeted at weights of morph targets has `SCALAR` accessor type, and the [current resample function](https://github.com/donmccurdy/glTF-Transform/blob/eae7adab2d221daa98e5d9cd6061ea1015d2b9b2/packages/functions/src/resample.ts#L106) takes this as the [element size](https://github.com/donmccurdy/glTF-Transform/blob/eae7adab2d221daa98e5d9cd6061ea1015d2b9b2/packages/core/src/properties/accessor.ts#L169) by [`getElement`](https://github.com/donmccurdy/glTF-Transform/blob/eae7adab2d221daa98e5d9cd6061ea1015d2b9b2/packages/core/src/properties/accessor.ts#L402), but it is [defined](https://github.com/KhronosGroup/glTF/blob/d202b79a816943beacdd9be6c936b63a1a9c0ea5/specification/2.0/Specification.adoc?plain=1#LL2581C1-L2581C138) as

> A morph target animation frame is defined by a sequence of scalars of length equal to the number of targets in the animated morph target.

So morph target animation with more than 1 targets would suffer from the issue referenced here.

Since the [new impl](https://github.com/donmccurdy/keyframe-resample-wasm/blob/d6896fa73af70370abf711486cdcc9c0b9b861ba/assembly/resample.ts#L23) does not use element size from accessor, the linked issue might be fixed by this.

kzhsw:
> There are several other WASM dependencies in the project now (Draco, Meshopt, MikkTSpace), so this module isn't necessarily the limiting factor on bundler and platform support.

The existing WASM dependencies are used by passing the wasm impl from options, that could make a difference for bundlers.

Also, should this be consdered as breaking change since this might require changes of some bundler users?

packages/functions/src/resample.ts
kzhsw: Perf note here: `input.clone()` would clone the underlying array, to avoid that:
```javascript
const array = input.getArray();
input.setArray(null);
const cloned = input.clone();
input.setArray(array);
```

Also, it should be possible to detect if `input` and `output` need to be cloned, by checking existance of other usages from `listParents`.

packages/functions/src/resample.ts
donmccurdy: Oh that's it, thank you! Found one other issue in the WASM implementation testing this, but it does look like this PR will now fix #290. 🎊

donmccurdy:
Here's where I ended up on backward-compatibility:

- The /functions module inlines the *JavaScript* implementation from the `keyframe-resample` module, which is still slow, but faster than the original implementation and should be fully backward compatible. No WASM binary is loaded by default.
- Users of the /functions module may optionally provide a WASM implementation from `keyframe-resample` or `keyframe-resample/compat` themselves. The latter is compatible with older bundlers, using a Data URI for the WASM build.
- The /cli module automatically uses the faster WASM implementation.

packages/functions/src/resample.ts
```diff
@@ -36,24 +57,54 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
...
+
+				if (samplerInterpolation === 'STEP' || samplerInterpolation === 'LINEAR') {
+					const input = sampler.getInput()!;
+					const output = sampler.getOutput()!;
+
+					accessorsVisited.add(input);
+					accessorsVisited.add(output);
+
+					const times = toFloat32Array(input.getArray()!, input.getComponentType(), input.getNormalized());
+					const values = toFloat32Array(output.getArray()!, input.getComponentType(), input.getNormalized());
```
kzhsw, line 82: If I was understanding corrently, this denormalized then input and output accessors, so should the cloned accessors be marked as `setNormalized(false)`, or should the denormalized times and values be normalized back?

packages/functions/src/resample.ts
```diff
@@ -70,92 +121,21 @@ export function resample(_options: ResampleOptions = RESAMPLE_DEFAULTS): Transfo
...
-	if (writeIndex !== input.getCount()) {
-		input.setArray(input.getArray()!.slice(0, writeIndex));
-		output.setArray(output.getArray()!.slice(0, writeIndex * output.getElementSize()));
-		sampler.setInput(input);
-		sampler.setOutput(output);
-	} else {
-		input.dispose();
-		output.dispose();
+	for (let i = 0; i < dstArray.length; i++) {
+		dstArray[i] = MathUtils.decodeNormalizedInt(dstArray[i], componentType);
```
kzhsw, line 137: Not benchmarked yet, but since denormalize is a pure algorithm, it might worth to implement it in wasm, sample [here](https://github.com/kzhsw/keyframe-resample-c/blob/81b04da50ba05128ac1b2a4fe0043d603a31178b/normalize.c#L481).

packages/functions/src/resample.ts
donmccurdy: Ah good call – let's normalize the values before writing back, yes.

packages/functions/src/resample.ts
donmccurdy: Maybe worth considering in the https://github.com/donmccurdy/keyframe-resample-wasm module — but I think I'm satisfied enough, as-is. I don't think I've ever seen a file using normalized accessors for animation, and it seems too early to add complexity to potentially optimize that.

PR #944, opened 2023-05-10
Add getTextureColorSpace(texture), fix inference on diffuse texture

### LINKED ISSUE #943: Wrong color space inference for spec/gloss diffuseTexture
Current code for inferring the color space of a texture, based on its assigned material slots:

https://github.com/donmccurdy/glTF-Transform/blob/ffd173382f3a5b6f8b0e676c6f8116910f9059fe/packages/cli/src/transforms/toktx.ts#L296

https://github.com/donmccurdy/glTF-Transform/blob/ffd173382f3a5b6f8b0e676c6f8116910f9059fe/packages/cli/src/transforms/ktxfix.ts#L28

Those patterns do not match the spec/gloss extension's "diffuseTexture", which is also sRGB. Perhaps this information should be stored on the graph edge.

donmccurdy:
- fixes #943 

Example:

```typescript
import { getTextureColorSpace } from '@gltf-transform/functions';

const baseColorTexture = material.getBaseColorTexture();
const normalTexture = material.getNormalTexture();

getTextureColorSpace(baseColorTexture); // → 'srgb'
getTextureColorSpace(normalTexture); // → null
```

### DIFF
--- a/packages/cli/src/transforms/ktxfix.ts
+++ b/packages/cli/src/transforms/ktxfix.ts
@@ -1,8 +1,6 @@
 import { KHR_DF_PRIMARIES_BT709, KHR_DF_PRIMARIES_UNSPECIFIED, read, write } from 'ktx-parse';
 import type { Document, Transform } from '@gltf-transform/core';
-import { MICROMATCH_OPTIONS } from '../util.js';
-import { listTextureSlots } from '@gltf-transform/functions';
-import micromatch from 'micromatch';
+import { getTextureColorSpace, listTextureSlots } from '@gltf-transform/functions';
 
 const NAME = 'ktxfix';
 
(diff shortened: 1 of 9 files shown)


donmccurdy:
@hybridherbst @marwie – this change allows extensions to declare the color space of a texture. Everything in glTF is 'srgb' or non-color data today, but I'm not sure whether that's true for your NEEDLE_lightmaps extension. Does this API seem OK for your use cases? 

```typescript
import { TextureChannel } from '@gltf-transform/core';

const { R, G, B, A } = TextureChannel;

// ...

public setLightmapTexture(texture: Texture | null): this {
    return this.setRef('lightmapTexture', texture, { channels: R | G | B, colorSpace: 'srgb-linear' });
}
```

Currently glTF Transform only uses this information to determine KTX2 compression settings. The color space string values come from [CSS Color Module Level 4](https://drafts.csswg.org/css-color/#predefined).

donmccurdy:
Perhaps it would be better to define only `isColor: true` rather than a specific color space, so the color space can later be inferred using more information, like the texture format. An emissive texture could be "srgb" or "srgb-linear" depending on the selected texture, if we are allowing (non-standard) OpenEXR textures. 🤔

donmccurdy:
Will go with `isColor: true`. It's a narrower claim that should be more forward-compatible with any extensions that might add linear, wide-gamut, or HDR color spaces.

PR #945, opened 2023-05-11
Improve performance of de-duplicating accessors with hash keys

### LINKED ISSUE #881: dedup: group accessors and images to accelerate comparsions
In function `dedupAccessors` and `dedupImages`, nested loops was used for checking duplication of accessors and textures, this should be fast when count of accessors or images are small, but slower when there are a large number of it.

Group accessors and images by a quick hash of some properties, then do the nested loop for each group. 
Maybe like the pesudo code here:
```typescript
function detectDuplicates(accessors: Accessor[]): Map<Accessor, Accessor> {
	const multiMap = new MultiValueMap<number, Accessor[]>;
	for (let i = 0; i < accessors.length; i++) {
		const a = accessors[i];
		const hashCode = makeHashCode(a.getType(), a.getComponentType(), a.getCount(), a.getNormalized(), a.getArray()!.byteLength);
		multiMap.add(hashCode, a);
	}
	const duplicateAccessors: Map<Accessor, Accessor> = new Map();
	multimap.values().forEach(accessors => {
		for (let i = 0; i < accessors.length; i++) {
			const a = accessors[i];
			const aData = BufferUtils.toView(a.getArray()!);

			if (duplicateAccessors.has(a)) continue;

			for (let j = 0; j < accessors.length; j++) {
				const b = accessors[j];

				if (a === b) continue;
				if (duplicateAccessors.has(b)) continue;

				if (a.getType() !== b.getType()) continue;
				if (a.getComponentType() !== b.getComponentType()) continue;
				if (a.getCount() !== b.getCount()) continue;
				if (a.getNormalized() !== b.getNormalized()) continue;
				if (BufferUtils.equals(aData, BufferUtils.toView(b.getArray()!))) {
					duplicateAccessors.set(b, a);
				}
			}
		}
	});

	return duplicateAccessors;
}
```

Keep existing mechanism and warn users with too many images or accessors in docs or in cli.

**Additional context**
screenshots about running `dedup` with a glb file with ~18k meshes:
The call tree view of chrome profiler
The source view of detectDuplicates after profiling

donmccurdy:
Testing against [Plant 3](https://sketchfab.com/3d-models/plant-3-5ecd9744ff6f4aa09766d796eec161ee), this change reduces time spent on the dedup() step by about 75% on my machine. I haven't seen any cases where comparing images is a bottleneck (except where the images themselves are large), so I'm waiting to make changes to that portion of the code.

- fixes #881

Thanks for the suggestion, @kzhsw!

### DIFF
--- a/packages/functions/src/dedup.ts
+++ b/packages/functions/src/dedup.ts
@@ -95,7 +95,7 @@ function dedupAccessors(document: Document): void {
 	}
 
 	// Add accessor to the appropriate hash group. Hashes are _non-unique_,
-	// intended only to reduce the set of comparisons required.
+	// intended to quickly compare everything accept the underlying array.
 	function hashAccessor(accessor: Accessor | null, group: Map<string, Set<Accessor>>): void {
 		if (!accessor) return;
 
@@ -125,11 +125,9 @@ function dedupAccessors(document: Document): void {
 
 				if (duplicates.has(b)) continue;
 
-				if (a.getType() !== b.getType()) continue;
-				if (a.getComponentType() !== b.getComponentType()) continue;
-				if (a.getCount() !== b.getCount()) continue;
-				if (a.getNormalized() !== b.getNormalized()) continue;
-				if (a.getSparse() !== b.getSparse()) continue;
+				// Just compare the arrays — everything else was covered by the
+				// hash. Comparing uint8 views is faster than comparing the
+				// original typed arrays.
 				if (BufferUtils.equals(aData, BufferUtils.toView(b.getArray()!))) {
 					duplicates.set(b, a);
 				}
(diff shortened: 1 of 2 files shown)


packages/functions/src/dedup.ts
```diff
@@ -73,105 +73,111 @@ export function dedup(_options: DedupOptions = DEDUP_DEFAULTS): Transform {
...
+	function hashAccessor(accessor: Accessor | null, group: Map<string, Set<Accessor>>): void {
+		if (!accessor) return;
+
+		const hash = [
+			accessor.getCount(),
+			accessor.getType(),
+			accessor.getComponentType(),
+			accessor.getNormalized(),
+			accessor.getSparse(),
+		].join(':');
```
kzhsw, line 108: Since the `join` here converts all to string, it might worth to encode numbers in hex or base64 to have shorter keys for memory effciency.

donmccurdy: It might, but I suspect it will be hard to accurately measure the time/memory tradeoff with JavaScript tooling here. We can always improve this further in other PRs, but I think I'd prefer to keep this simple for now.

Issue #961, opened 2023-05-24
Not all points are transformed if primitives don't share the same vertex accessors


alanb-sony:
If a mesh is made up of multiple primitives and the vertex accessors aren't the same then when calling `transformMesh` a single `skipIndices` set is used for all primitives resulting in some vertices not being transformed as they have the same indexes but in different accessors.
In [models.zip](https://github.com/donmccurdy/glTF-Transform/files/11553653/models.zip) `single_vertex_array.glb` is transformed correctly but `separate_vertex_array.glb` only the first triangle is transformed.

Demo code:
```
const io = new NodeIO();
const source = await fs.promises.readFile('separate_vertex_array.glb');
const model = await io.readBinary(source);
const quaternion = quaternionFromEulerYXZ(pitch, yaw, roll);
const matrix: mat4 = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
GLMat4.fromQuat(matrix, quaternion);
GLMat4.scale(matrix, matrix, [scale, scale, scale]);
for (const mesh of model.getRoot().listMeshes()) {
  transformMesh(mesh, matrix);
}
```

donmccurdy:
Thank you! Definitely a bug, the different possible combinations of shared/unshared vertex attributes are more complex than I fully thought through when implementing this.

> ... they have the same indexes but in different accessors.

Could you say more about the structure, if possible? Do the primitives not share vertex attributes at all, or only a subset? If no vertex attributes a shared, a workaround for now might be to iterate over the primitives and call [`transformPrimitive`](https://gltf-transform.donmccurdy.com/modules/functions/functions/transformPrimitive) instead.

alanb-sony:
Just calling `transformPrimitive` would work in this case but would break in the case where the primitives do share vertices (my application generates models with all the vertices in a single accessor which works fine but users can upload their own models, some of which have this oddity of using separate vertices for each primitive). In my particular use case there typically aren't tangents or normals generated so it's only positions I'm worried about but of course it'd be nice to fix this for the general case.

donmccurdy:
To be fixed by #987.

alanb-sony:
thanks, confirmed fixed in my code

PR #962, opened 2023-05-24
fix for transforming meshes with non-shared indices

alanb-sony:
fix for #961
not sure that this is the right approach but does fix my test case


donmccurdy:
Thank you! Let me study this one a bit. I think I've made incorrect assumptions earlier in the function. We will need to check for shared normals and tangents as well, since those are also potentially transformed.

donmccurdy:
Thanks for calling this issue out @alanb-sony! I've fixed this is a similar way to your PR and will merge that in #987.

PR #1026, opened 2023-07-14
Fix parent/child validation regression

### LINKED ISSUE #1025: Breaking change to addChild between 3.0 and 3.1
**Describe the bug**
In v3.1.0+ using `addChild` to change the parent of a node from the scene to another node no longer seems to work. The node will remain a child of the scene after the `addChild` call.

**To Reproduce**
The following code will print `1` on `3.0.0` and `2` on `3.1.0`.

```js
import { NodeIO } from "@gltf-transform/core";

(async () => {
  const io = new NodeIO();
  const doc = await io.read("DamagedHelmet.glb");

  const newRootNode = doc.createNode("test");
  const scene = doc.getRoot().getDefaultScene();
  const childNode = scene.listChildren()[0];
  scene.addChild(newRootNode);
  newRootNode.addChild(childNode);

  console.log(scene.listChildren().length);
})();
```

The behavior can be made identical by adding `scene.removeChild(childNode)` to the above snippet.

**Expected behavior**
I would expect there to only be one child node to the scene in both versions.

**Versions:**
 - Version: 3.1.0+
 - Environment: Node.js

donmccurdy:
Requirements:

1. Nodes MAY be root children of multiple Scenes
2. Nodes MUST NOT be children of >1 Node
3. Nodes MUST NOT be children of both Nodes and Scenes

- Fixes #1025
- Fixes regression from #833

### DIFF
--- a/packages/core/src/properties/node.ts
+++ b/packages/core/src/properties/node.ts
@@ -7,6 +7,7 @@ import { ExtensibleProperty, IExtensibleProperty } from './extensible-property.j
 import type { Mesh } from './mesh.js';
 import { COPY_IDENTITY } from './property.js';
 import type { Skin } from './skin.js';
+import type { Scene } from './scene.js';
 
 interface INode extends IExtensibleProperty {
 	translation: vec3;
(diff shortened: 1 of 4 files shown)


packages/core/src/properties/node.ts
```diff
@@ -176,29 +187,50 @@ export class Node extends ExtensibleProperty<INode> {
...
 	public addChild(child: Node): this {
-		// Remove existing parent.
+		// Remove existing parents.
 		if (child._parentNode) child._parentNode.removeChild(child);
+		if (child._parentScenes.size) {
+			const prevSize = child._parentScenes.size;
+			for (const scene of child._parentScenes) {
+				scene.removeChild(child);
+			}
+			console.log(`parentScenes: ${prevSize} -> ${child._parentScenes.size}`);
```
donmccurdy, line 213: Removed logging in https://github.com/donmccurdy/glTF-Transform/commit/2cfb488ae5ef76d370867e52f2403fc3d4e4e2ac.

PR #1029, opened 2023-07-16
Fix attribute dynamic range

rotu:
Previously, when computing a position attribute's tolerance, we were subtracting the maximum coordinate from the minimum coordinate (so the points [0,0,99] and [0,0,100] would use range=(100-0)=100 instead of range=(100-99)=1 in welding tolerance computations.

### DIFF
--- a/packages/functions/src/weld.ts
+++ b/packages/functions/src/weld.ts
@@ -336,7 +336,8 @@ function getAttributeTolerance(semantic: string, attribute: Accessor, tolerance:
 	_a.length = _b.length = 0;
 	attribute.getMinNormalized(_a);
 	attribute.getMaxNormalized(_b);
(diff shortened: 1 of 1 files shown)


rotu:
A consequence of the underlying issue is that position tolerances previously had to be tighter than necessary to achieve good results with off-origin meshes. Not sure how the thresholds were chosen, but DEFAULT might now be reduced.

https://github.com/donmccurdy/glTF-Transform/blob/f2c556b055d2e4701ab6566e89e3ae6682ef2d19/packages/functions/src/weld.ts#L47-L54

donmccurdy:
Thanks @rotu! Agreed on the change. The default tolerance here is arbitrary, but intended to complement the `quantize()` function by being somewhere in the neighborhood of 14-bit precision. I don't think we'll need to change that part, or at least not as a result of this PR.

PR #1035, opened 2023-07-21
Reduce normal tolerance for Weld

### LINKED ISSUE #1028: weld has huge normal tolerance
`gltf-transform weld` welds vertices with relatively large discrepancies in normal. There is not a command-line option to change this threshold.

image
image

rotu:
Fix #1028
~~Chose 0.01 as a default tolerance which corresponds to ±arcsin(0.01)/2 = ±0.286°.~~

Chose 0.05 as a default tolerance which corresponds to ±arcsin(0.05)/2 = ±1.43°.

### DIFF
--- a/packages/functions/src/weld.ts
+++ b/packages/functions/src/weld.ts
@@ -48,7 +48,7 @@ const Tolerance = {
 	DEFAULT: 0.0001,
 	TEXCOORD: 0.0001, // [0, 1]
 	COLOR: 0.01, // [0, 1]
-	NORMAL: 0.01, // [-1, 1]
+	NORMAL: 0.05, // [-1, 1]
 	JOINTS: 0.0, // [0, ∞]
 	WEIGHTS: 0.01, // [0, ∞]
 };
--- a/packages/functions/src/weld.ts
+++ b/packages/functions/src/weld.ts
@@ -48,7 +48,7 @@ const Tolerance = {
 	DEFAULT: 0.0001,
 	TEXCOORD: 0.0001, // [0, 1]
 	COLOR: 0.01, // [0, 1]
-	NORMAL: 0.5, // [-1, 1]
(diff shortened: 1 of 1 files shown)


rotu:
Oops! I had forgotten to submit this PR.

I think it's a pretty good threshold, but I'm not sure if there's a more principled way to set these values!

donmccurdy:
Would you be able to share a model that reproduces the bug described in #1028? 

One test case I've used before is this:

https://sketchfab.com/3d-models/sculpt-january-day-19-lovecraftian-34ad2501108e4fceb9394f5b816b9f42

```bash
gltf-transform optimize in.glb out.glb --verbose
```

The largest mesh in that scene has just under 900K vertices, and the normal tolerance makes a big difference when welding them:

- before: 873,244 → 228,453 (–73.84%) vertices
- after: 873,244 → 802,173 (–8.14%) vertices

Aggressive welding is also an important step before simplifying a model. I'm hoping there's a "happy medium" value that can give us a better default here, but perhaps this option just needs to be exposed in the CLI too.

rotu:
https://wormhole.app/YO9Mr#-YSuDqBOBOySwiBYmNR7tA

The visible weld issue on my model goes away at a normal tolerance of about 0.3. So maybe 0.1 is a good threshold?

rotu:
Here's a pathological example that's wrecked by `optimize`, though not by the `weld` step alone: https://sketchfab.com/3d-models/scifi-hexsphere-cb364832b9994b768dba6245e6b3f51b

donmccurdy:
Thanks! I think 0.05 seems about right. I'd recommend higher tolerance, closer to 0.5, when `weld()` is used as a pre-process before simplification for LODs. Simplification (for now, at least) cannot do much with disjoint topology.

PR #1039, opened 2023-07-24
Use teapot emoji to mean "geometry"

rotu:
🫖 is a better choice than 🕋 to mean "geometry". It's already a [commonly recognized representation of 3D geometry](https://en.wikipedia.org/wiki/Utah_teapot). And the Kaaba is a bit culturally loaded.


donmccurdy:
Good call -- thanks!

donmccurdy:
Aside: only `cli.ts` needs to be updated, the .md file will update automatically and the .txt file is outdated, I'll remove that.

Issue #1058, opened 2023-08-07
Improve weld to not remove empty nodes when tolerance is 0


ReedCopsey:
When running weld, including indirectly via functions such as draco(), prune is called.

This causes empty nodes to be pruned, even if there is no technical reason to do so.

**Describe the solution you'd like**
Ideally, when weld is called with a tolerance of 0, the prune call would be skipped, as weld could never have created an empty mesh.

**Additional context**
See [discussion here for details](https://github.com/donmccurdy/glTF-Transform/discussions/1050) and inspiration.

PR #1059, opened 2023-08-07
Add a check to avoid prune in weld when tolerance is zero

ReedCopsey:
Addresses feature request in #1058. Since weld with a tolerance of 0 will never remove a mesh, this avoids the call to prune in that case.

This prevents draco compression from stripping empty nodes in a gltf.


donmccurdy:
Thanks @ReedCopsey! This change is now published as v3.5.1.

ReedCopsey:
Thank you @donmccurdy. Updated and it's working perfectly now.

Issue #1116, opened 2023-10-04
Weld() error when gltf contains a point primitive


Gmadges:
**Describe the bug**
When a Document contains a point primitive that is not indexed, the weld() function errors with a 
`error: Cannot read properties of null (reading 'getCount')`

**Versions:**
 - Version: v3.6.0
 - Environment: Node.js

**Additional context**

donmccurdy:
Fixed by #1117, thanks!

PR #1117, opened 2023-10-04
Fix for Weld() Error

Gmadges:
#1116

There is a code path that results in `error: Cannot read properties of null (reading 'getCount')`

Where a non-indexed point primitive exits early from weldPrimitive(), but then fails the `prim.getIndices()!.getCount() === 0` check because `prim.getIndices() === null`

I've added a really simple fix here, but it's maybe not the cleanest an alternative could be to perform a specific point prim check instead and then continue


donmccurdy:
Thanks @Gmadges!

Issue #1133, opened 2023-10-24
Performance improvements for large, complex models


donmccurdy:
Performance degrades when optimizing large, complex models. Particularly those with many tens or hundreds of thousands of properties like accessors.

1. Disposal can be very slow, particularly Root->Accessor and Root->Node ref lists.
    - Related: https://github.com/donmccurdy/glTF-Transform/issues/954
    - Functions: flatten, prune, join
2. setElement / getElement are convenient, but slow for bulk operations
    - Goal: Use faster alternatives for most internal operations
    - Functions: weld, join
3. Cloning a large Document (100K nodes, 300K graph edges): ~1000ms in v3
4. Reading a large Document (100K nodes, 300K graph edges): ~700ms in v3

There might be more efficient ways to represent the property graph structure. I'm wondering, in particular, if edges could be represented without separate objects for each.

It'd be very helpful to create some benchmarks that run in CI. Consider:

- https://github.com/benchmark-action/github-action-benchmark

Makio64:
Weld is actually crashing my app on large mesh maybe using a subdivision strategy will also speed up the weld process like quadtree grid / bvh ?

donmccurdy:
Possibly! There's a separate issue specifically for `weld()` at https://github.com/donmccurdy/glTF-Transform/issues/738 ... it's historically been difficult. I think there's a hard limit somewhere around 16M vertices based on limits of JavaScript's `Map` object.

donmccurdy:
Related:

- https://github.com/donmccurdy/property-graph/issues/121

donmccurdy:
@Makio64 if you'd be interested in trying it, I've made an alpha release of glTF Transform v4 with performance improvements to `weld()` from #1237. To test:

```bash
npm install --global @gltf-transform/cli@next

gltf-transform optimize input.glb output.glb
```

donmccurdy:
@Makio64 yes! You'd need to install from the `@next` tag for whichever glTF Transform packages you're using. For example:

```
npm install --save @gltf-transform/core@next @gltf-transform/extensions@next @gltf-transform/functions@next
```

Makio64:
Just tested it : I didnt do a proper benchmark but it's looks much faster on many models which previously felt very slow. 

Well done!

donmccurdy:
Remaining work tracked in https://github.com/donmccurdy/glTF-Transform/issues/1315.

PR #1141, opened 2023-10-25
v4: Upgrade to property-graph v2 (breaking change)

### LINKED ISSUE #954: Support formal Ref/RefList/RefMap types
Related:

- https://github.com/donmccurdy/property-graph/issues/20

glTF Transform v4 might be the time to make this change, releasing a new version of `property-graph` at the same time. This should significantly increase the performance of `prune` and `dedup` operations in large files.

donmccurdy:
Changes:

- Upgrade to property-graph@v2
- Fixes #954 
- Progress on #1133 
- 🚀 Improved performance when processing large, complex scenes
- ‼️ **Breaking change for authors of extensions**. One-to-many references now require use of the RefSet, RefList, or RefMap types. See updates to core classes and extensions in this PR.


donmccurdy:
Note - the only bench mark that really improves with this PR is 'flatten.bench.ts', but real-world testing on larger scenes like PlantComplex showed large improvements on join and dedup operations as well. Mainly this allows faster disposal of objects, which was previously not well optimized for scenes with many nodes or accessors.

PR #1167, opened 2023-11-13
Docs Fix: Update prune docs to hint at options instead of claiming there are none

johannesvollmer:
thanks for this library :)


donmccurdy:
Thank you @johannesvollmer!

PR #1186, opened 2023-11-29
Draft: Remove side effects from quantize()

### LINKED ISSUE #1185: Disable `dedup` and `prune` in `quantize` function
Some models provided by designers need to maintain their original structure, but `dedup` and `prune` functions will destroy the structure of the glTF.

Add options to disable `prune` and `dedup` functions.

Add flags to modified nodes and materials, `prune` and `dedup` will only destroy the modified data.

**Additional context**
The real-world scenario is that designers have a series of models, and we optimize the material size through quantization. These materials of glTF have the same properties in glTF, so they are removed. However, we actually need to adjust these materials at runtime to achieve different effects in different locations.

So I don't want to run `dedup` function in my glTF process pipeline.

donmccurdy:
- fixes #1185 

Still a draft. So far this PR tracks which properties have been created or cloned during the quantization process, and considers only those properties for pruning.

We'll still need to consider how to implement deduplication without actually running the whole `dedup()` transform. Currently the [implementation of `dedup`](https://github.com/donmccurdy/glTF-Transform/blob/main/packages/functions/src/dedup.ts)  is fairly complicated — it was written before `a.equals(b)` was available I think. Maybe it could be simpler by using that. But we need to be pretty careful to not introduce performance regressions for models that may have many thousands of accessors. My rough ideas for next steps would be:

1. expose separate functions like
  - `getDuplicateMaterials(materials)`
  - `getDuplicateAccessors(accessors)`
  - ...
2. call those functions with all materials/accessors/... in `dedup()` and only with modified properties in `quantize()`
3. clean things up and simplify as much as possible

If anyone is interested in improving `dedup()` in a separate PR please let me know, I'm not sure yet when I'll get to that.

### DIFF
--- a/packages/functions/src/quantize.ts
+++ b/packages/functions/src/quantize.ts
@@ -11,19 +11,16 @@ import {
 	Primitive,
 	PrimitiveTarget,
 	Property,
-	PropertyType,
 	Skin,
 	Transform,
 	vec2,
 	vec3,
 	vec4,
 } from '@gltf-transform/core';
-import { dedup } from './dedup.js';
 import { fromRotationTranslationScale, fromScaling, invert, multiply as multiplyMat4 } from 'gl-matrix/mat4';
 import { max, min, scale, transformMat4 } from 'gl-matrix/vec3';
 import { InstancedMesh, KHRMeshQuantization } from '@gltf-transform/extensions';
 import type { Volume } from '@gltf-transform/extensions';
-import { prune } from './prune.js';
 import { createTransform } from './utils.js';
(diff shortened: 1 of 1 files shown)


donmccurdy:
I think https://github.com/donmccurdy/glTF-Transform/pull/1188 mostly removes the need for this PR, but will take a closer look to see if `dedup()` can be cleaned up a bit anyway.

donmccurdy:
See also, https://github.com/donmccurdy/glTF-Transform/pull/1300.

PR #1235, opened 2024-01-17
Update documentation for Accessor usage

harrycollin:
The usage stated that you can call a non-existent method on Document called listBuffers(). This exists on Root instead. 

Changed:
   .setBuffer(doc.listBuffers()[0]);

To: 
   .setBuffer(doc.getRoot().listBuffers()[0]);


donmccurdy:
Thank you!

Issue #1136, opened 2023-10-25
🛣️ v4 Roadmap 


donmccurdy:
## v4 Roadmap

_Roadmap and priorities for glTF Transform v4.x releases, thanks to project sponsors and contributors._

### Maintenance (Ongoing)

- **Improve `/cli` code health and performance:** The dependency currently used for the glTF Transform CLI (Caporal.js) has limited maintenance recently. I’ve evaluated alternatives and I still prefer Caporal.js, but the `cli` module is currently higher cost in terms of maintenance time, and I’d like to improve that.
- **Website & documentation:** The website (https://gltf-transform.dev/) focuses mostly on API documentation. I’d like for the website to also provide tutorials for common use cases, and more compelling illustrations and demos of how the glTF Transform project can be used.
- **Performance:** Improving processing speed, particularly for large and complex scenes

### v4.x (features)

- `/functions`
    - #802 
- `/extensions`
    - #1134 
    - #1135 
    - Improvements to Extension APIs
- `/cli`
    - Support for upcoming KTX Software release
- … other small tasks from [[v4 Milestone](https://github.com/donmccurdy/glTF-Transform/milestone/28)](https://github.com/donmccurdy/glTF-Transform/milestone/28)

### ****v4.0 (breaking changes)****

- Performance optimization 🚀
    - Optimize `/core` APIs for faster processing of large files
    - Optimize `/extension` APIs for faster processing of large files

> [!NOTE]
> If you have questions or ideas you’d prefer to share privately, please reach out to me at `dm@donmccurdy.com`.

> [!WARNING]
> Please don’t request new features in this thread — open a new issue instead. Thanks!

PR #1268, opened 2024-02-16
fix(simplify): Correct usage of 'target_index_count'

### LINKED ISSUE #1263: Simplify breaks due to assertion failed
**Describe the bug**

I'm running into a bug where `gltf-transform simplify` from the command line breaks with an assertion failed error.

Upon closer examination, it looks like this error is coming from this line here (from the meshopt simplify function):
https://github.com/zeux/meshoptimizer/blob/master/js/meshopt_simplifier.js#L178 , where `targetCount > indices.Length` after one iteration therefore failing the assertion.

One thing I did notice from the example of simplify here: https://github.com/zeux/meshoptimizer/blob/master/demo/simplify.html -- it looks like the `targetCount` is calculated based on index count and not the vertex position count as used by gltf-transform. I'm wondering if this is intended, and whether that relates to the bug?

**To Reproduce**
I'm unable to share the model I was testing with, but I'll see if I can recreate the bug with another model.

**Versions:**
 - Version: `cli.js 4.0.0-alpha.5 — Command-line interface (CLI) for the glTF Transform SDK.`
 - Environment: Command line

Thanks!

donmccurdy:
- fixes #1263

### DIFF
--- a/packages/functions/src/simplify.ts
+++ b/packages/functions/src/simplify.ts
@@ -129,6 +129,7 @@ export function simplifyPrimitive(document: Document, prim: Primitive, _options:
 	const logger = document.getLogger();
 	const position = prim.getAttribute('POSITION')!;
 	const srcIndices = prim.getIndices()!;
+	const srcIndexCount = srcIndices.getCount();
 	const srcVertexCount = position.getCount();
 
 	let positionArray = position.getArray()!;
@@ -159,7 +160,7 @@ export function simplifyPrimitive(document: Document, prim: Primitive, _options:
 
 	// (2) Run simplification.
 
-	const targetCount = Math.floor((options.ratio * srcVertexCount) / 3) * 3;
+	const targetCount = Math.floor((options.ratio * srcIndexCount) / 3) * 3;
 	const [dstIndicesArray, error] = simplifier.simplify(
 		indicesArray as Uint32Array,
 		positionArray as Float32Array,
--- a/packages/functions/test/simplify.test.ts
+++ b/packages/functions/test/simplify.test.ts
@@ -3,7 +3,7 @@ import path, { dirname } from 'path';
 import { getBounds, Document, NodeIO, Primitive } from '@gltf-transform/core';
 import { KHRDracoMeshCompression, KHRMeshQuantization } from '@gltf-transform/extensions';
 import { weld, unweld, simplify } from '@gltf-transform/functions';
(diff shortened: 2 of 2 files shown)


donmccurdy:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1268?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

Current dependencies on/for this PR:
* `main`
  * **PR #1267** <a href="https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1267?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1268** <a href="https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1268?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1268?utm_source=stack-comment).

donmccurdy:
@marwie very likely this would fix the discrepancy in the ratio (50% vs 90%), yes. I think it is unrelated to the error metric, though.

marwie:
Thank you. I've updated to latest 3 release today but perhaps it's worth updating to 4 already then. From the roadmap it seems like there are no breaking changes https://github.com/donmccurdy/glTF-Transform/issues/1136

donmccurdy:
I haven't written the changelog for v4 yet, but so far the breaking changes are minimal. Some options renamed, or with different default values. If you're using TypeScript, it should catch things. The larger breaking change in v4 will be #1141, which requires a (hopefully small) change to extension implementations. See the changes to EXT_mesh_gpu_instancing in that PR, for an example.

This fix could be patched to a v3.10.x release safely, if that's better.

donmccurdy:
Published to published v3.10.x. ✅ 

#1141 is not on the alpha release yet — everything on the alpha release has already been merged to the `main` branch. See https://github.com/donmccurdy/glTF-Transform/milestone/28 for the full list of what's in the alpha vs. still planned.

marwie:
Thanks a lot! Time to update then 😎

PR #1279, opened 2024-02-21
Optional option to join to suppress the cleanup step.

### LINKED ISSUE #1278: Join also does a prune inside, with settings that are overly aggressive for certain use cases.
I have a workflow where I am using empty objects with a custom attribute as placeholders
that get loaded and replaced at run time.
prune clears them because they end up being empty leaves.

I would like an option to suppress the clean up step (or override the prune properties)

I tried making my own local version of join, but had dependency issues.
I also tried detaching and keeping a reference to the empties, and then re-attaching them at the end,
but they are disposed and I can't find a way to keep them alive through all the transforms.

**Additional context**
I am also including a PR to add the functionality.  It's small, self contained, documented, and doesn't change any default behavior.

Archimagus:
Resolves #1278

It is too aggressive, this gives the option to suppress it allowing for finer control, while not affecting the default behavior.

Also added option to skip cleanup on simplify and weld.

### DIFF
--- a/packages/functions/src/join.ts
+++ b/packages/functions/src/join.ts
@@ -44,6 +44,10 @@ export interface JoinOptions {
 	 * have no effect. Default: false.
 	 */
 	keepNamed: boolean;
+	/**
+	 * Prevents the final {@link prune} step from being triggered.
+	 */
+	suppressCleanup?:boolean;
 }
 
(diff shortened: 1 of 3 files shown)


donmccurdy:
Superceded by https://github.com/donmccurdy/glTF-Transform/pull/1300, thank you for raising the issue!

PR #1280, opened 2024-02-21
feat(core): Add io.jsonToBinary()

### LINKED ISSUE #746: Add JSONToBinary method
Apart from natural symmetry to the existing `binaryToJSON` method, I want to build a manifoldness extension which uses sparse accessors. Since this library doesn't support writing them, I'd like to use it to make everything else, write the JSON, edit it a bit to add a sparse accessor, then convert to binary, where this function would be helpful. If I read the JSON back, I'll lose the sparse accessor information. 

I believe just the `JSONToBinary` function is needed for this workflow. I can simply create a standard accessor for the sparse data to ensure its `bufferView` is written to the output, then reference its index in my JSON modification. 

Adding sparse accessor write support to the library would also work, but I can see why that was avoided as it complicates the API. Or I can just write the GLB myself, but I'd much rather not duplicate all the nice work you've done here. 

**Additional context**
I'm looking to make a lossless round-trip between glTF and my `manifold` library of Boolean mesh operations. If you join two meshes, they must still be separate draw calls (primitives) since they have different properties/textures. This requires the intersection vertices to be duplicated, since a single position has different vertex properties on the two sides. That in turn breaks manifoldness, as those vertices are no longer associated. It cannot be reliably fixed by merging identical verts, as manifoldness can also require multiple verts at the same position that are not attached topologically. 

To fix this, I want to propose an extension where a single object is represented as a single mesh of multiple primitives, but an alternative single primitive is also available that references the same data but uses a sparse accessor on the indices to merge them. This primitive will not refer to mesh properties beyond position. In this way, a renderer has access to the standard set of primitives with properties, while a CAD app can access the manifold verts without extra properties and have the relation between the two without duplication.

### LINKED ISSUE #1171: Cant convert to glb
**Describe the bug**
On this particular model I cant convert the gltf to glb. but it works perfectly using the gltf-pipeline 
It says there is no buffer but I dont understand whats wrong, works fine with other model i tested.

**To Reproduce**
here the model : https://www.dropbox.com/scl/fi/calx036yfn0h3qt9x67vw/camera_curve.gltf?rlkey=vltp12bbg7s021wmur0wa2kjp&dl=0
here my code for converting : 
```
let io = null
const initNodeIO = async () => {
	let dracoDecoder = await draco3d.createDecoderModule()
	let dracoEncoder = await draco3d.createEncoderModule()
	io = new NodeIO()
		.registerExtensions(ALL_EXTENSIONS)
		.registerDependencies({
			'meshopt.decoder': MeshoptDecoder,
			'meshopt.encoder': MeshoptEncoder,
		})
		.registerExtensions(KHRONOS_EXTENSIONS)
		.registerDependencies({
			'draco3d.decoder': dracoDecoder,
			'draco3d.encoder': dracoEncoder,
		})
}
initNodeIO 

async function convertGltfToGlb(inputPath) {
	const document = await io.read(inputPath)

	// write the glb
	const outputPath = inputPath.replace(path.extname(inputPath), '.glb')
	await io.write(outputPath, document)

	console.log(`Converted ${inputPath} to ${outputPath}`)
	return outputPath
}
```

**Expected behavior**
Convert the model with no problem.

**Versions:**
 - Version: 3.9.0
 - Environment: Nodejs Windows & Mac

donmccurdy:
Example:

```javascript
const { json, resources } = await io.writeJSON(document);
const glb = await io.jsonToBinary({ json, resources });
```

Related:

- Fixes #746
- Fixes #1171

Remaining:

 - [ ] More tests

### DIFF
--- a/packages/core/src/io/node-io.ts
+++ b/packages/core/src/io/node-io.ts
@@ -172,7 +172,7 @@ export class NodeIO extends PlatformIO {
 
 	/** @internal */
 	private async _writeGLB(uri: string, doc: Document): Promise<void> {
-		const buffer = Buffer.from(await this.writeBinary(doc));
+		const buffer = await this.writeBinary(doc);
 		await this._fs.writeFile(uri, buffer);
 		this.lastWriteBytes = buffer.byteLength;
 	}
--- a/packages/core/src/io/platform-io.ts
+++ b/packages/core/src/io/platform-io.ts
@@ -124,65 +124,75 @@ export abstract class PlatformIO {
 		jsonDoc = this._copyJSON(jsonDoc);
 		this._readResourcesInternal(jsonDoc);
 		this._assertResources(jsonDoc);
+
 		const { json, resources } = jsonDoc;
 
 		if (!json.buffers || json.buffers.length <= 1) {
 			return this._jsonToBinary(jsonDoc);
 		}
 
-		const buffers = json.buffers || [];
+		const buffers = json.buffers;
 		const bufferViews = json.bufferViews || [];
+
+		const glbBufferIndex = buffers.findIndex(({ uri }) => !!uri);
+		const glbBufferResources: Uint8Array[] = [];
+
+		let byteOffset = 0;
 		const bufferByteOffsets = new Uint32Array(buffers.length);
 
-		const defaultBufferIndex = buffers.findIndex(({ uri }) => !!uri);
-		const defaultBufferResources = [] as Uint8Array[];
-
-		let nextByteOffset = 0;
-
+		// Iterate over buffers, adding everything with a URI to the GLB buffer,
+		// and ignoring fallback buffers without URIs.
 		for (let bufferIndex = 0; bufferIndex < buffers.length; bufferIndex++) {
 			const buffer = buffers[bufferIndex];
-			if (!buffer.uri) continue; // fallback
 
-			bufferByteOffsets[bufferIndex] = nextByteOffset;
-			defaultBufferResources.push(resources[buffer.uri!]);
-			delete resources[buffer.uri!];
+			if (!buffer.uri) {
+				continue; // fallback buffer
+			}
 
-			nextByteOffset += buffer.byteLength;
+			bufferByteOffsets[bufferIndex] = byteOffset;
+			glbBufferResources.push(resources[buffer.uri]);
+			delete resources[buffer.uri];
+
+			byteOffset += buffer.byteLength;
 		}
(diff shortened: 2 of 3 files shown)


donmccurdy:
Current dependencies on/for this PR:
* **#1280** <a href="https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1280?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1280?utm_source=stack-comment).

donmccurdy:
Testing on meshopt-compressed inputs with the script below:

```js
import { NodeIO } from '@gltf-transform/core';
import { ALL_EXTENSIONS } from '@gltf-transform/extensions';
import { MeshoptDecoder } from 'meshoptimizer';
import fs from 'fs/promises';

await MeshoptDecoder.ready;

const io = new NodeIO()
  .registerExtensions(ALL_EXTENSIONS)
  .registerDependencies({ 'meshopt.decoder': MeshoptDecoder });

const jsonDoc = await io.readAsJSON('/Users/donmccurdy//Desktop/bluesword-meshopt.gltf');

console.log(JSON.stringify(jsonDoc.json.bufferViews, null, 2));
console.log('-------------');

const glb = await io.jsonToBinary(jsonDoc);

const jsonDoc2 = await io.binaryToJSON(glb);

console.log(JSON.stringify(jsonDoc2.json.bufferViews, null, 2));

await fs.writeFile('/Users/donmccurdy/Desktop/bluesword-packed.glb', glb);
```

donmccurdy:
I've spent a while trying to provide this API...

```js
const glb = await io.jsonToBinary(jsonDoc);
```

... but a fundamental problem with this approach seems to be, that if we don't parse the JSON into a [Document](https://gltf-transform.dev/modules/core/classes/Document), which I'd _expect_ a method of that name to avoid, then extensions are never initialized and we have no way to understand extensions' references to buffers. And the buffers need to be repacked when writing a multi-buffer input to GLB. This shows up with Meshopt compression, but other extensions could theoretically do the same thing.

For now the best option remains to read the document, unpartition its buffers, and then write to GLB:

```js
import { unpartition } from '@gltf-transform/functions';
const document = io.readJSON(jsonDoc);
await document.transform(unpartition());
const glb = await io.writeBinary(document);
```

PR #1298, opened 2024-03-06
Added prune options in optimize command

subho57:
Discussion Reference: https://github.com/donmccurdy/glTF-Transform/discussions/1296


donmccurdy:
Thanks @subho57! I think I will remove `--prune-indices`, since I expect that to be very rarely used if at all, but let me know if you have any concerns with that. :)

PR #1300, opened 2024-03-07
feat(functions): Provide option to skip cleanup in transforms

### LINKED ISSUE #1185: Disable `dedup` and `prune` in `quantize` function

### LINKED ISSUE #1278: Join also does a prune inside, with settings that are overly aggressive for certain use cases.

donmccurdy:
Some workflows require preserving seemingly unused or duplicate resources. Many transforms currently remove those resources as a side effect of cleaning up temporary resources created during normal operations. This PR provides an opt-out, `cleanup: false`, for these functions. Users opting out of cleanup will likely want to run `prune()` and `dedup()` manually, with options set for their needs, later in their pipelines.

Eventually I'd like to change the implementations of these functions so that they can clean up temporary resources without these side effects, but that will be a larger task for a future PR.

Related:

- fixes #1278
- fixes #1185
- alternative #1279
- alternative #1186
- related #1281

Remaining:

- [x] more unit tests

### DIFF
--- a/packages/functions/src/flatten.ts
+++ b/packages/functions/src/flatten.ts
@@ -6,10 +6,20 @@ import { createTransform } from './utils.js';
 const NAME = 'flatten';
 
 /** Options for the {@link flatten} function. */
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface FlattenOptions {}
+export interface FlattenOptions {
+	/**
+	 * Whether to perform cleanup steps after completing the operation. Recommended, and enabled by
+	 * default. Cleanup removes temporary resources created during the operation, but may also remove
+	 * pre-existing unused or duplicate resources in the {@link Document}. Applications that require
+	 * keeping these resources may need to disable cleanup, instead calling {@link dedup} and
+	 * {@link prune} manually (with customized options) later in the processing pipeline.
+	 */
+	cleanup?: boolean;
+}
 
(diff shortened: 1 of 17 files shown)


donmccurdy:
* **#1300** <a href="https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1300?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #1340, opened 2024-04-02
chore(functions): Expose getBounds from /functions

donmccurdy:
Less confusing this way, no one expects to find it in /core. We do still
need to use it in /extensions, so it's implemented in /core, exported
with a hidden tag, and then publicly exported from /functions.


donmccurdy:
* **#1340** <a href="https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1340?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Issue #1315, opened 2024-03-16
Optimize inspect(), weld(), and join() for scenes with many primitives sharing vertex streams


donmccurdy:
For models reusing the same vertex attributes heavily, parsing and welding are slower than they ideally would be. The attached sample, from a recent [three.js discussion](https://discourse.threejs.org/t/excessive-memory-usage-for-gltf-file/62807), has 10,000 mesh primitives, all sharing the same vertex position and normal attributes, but with different indices. It's not an optimized file – the primitives should be merged — but still, I think that glTF Transform can and should process the file more quickly than it currently does.

When running a `weld`, I think the issue is that each primitive is handled independently. With `inspect`, I'm not sure where the cost lies.

[oval.gltf.zip](https://github.com/donmccurdy/glTF-Transform/files/14622204/oval.gltf.zip)

donmccurdy:
Essentially the same issue coming up for `join()` in #1317, which I'll merge into this issue. The functions need to either be smarter about working only with the specific vertices indexed by a given primitive (not its entire vertex stream, possibly shared), or the functions need to be refactored to operate on some concept of a vertex stream. I suspect the first is more maintainable.

Another example file, available at:

https://www.dropbox.com/scl/fi/q1uisxtqoe7hj4d5gzm1r/world.zip?rlkey=1gt9q5t4ns9v8xwwx8gy6geo2&dl=0

donmccurdy:
_Concluding notes —_

***

## dequantize ✅

OK to iterate vertex attributes, because the attributes are processed only once, never cloned.

## reorder ✅

Calls remapAttribute, iterates over the source attribute. Appears not to write any more than necessary, either way. Improved by switching from remapAttribute to compactAttribute. Further improved by reducing excess iteration over primitives sharing the same vertex stream.

- https://github.com/donmccurdy/glTF-Transform/pull/1358

## unweld ✅

Did a clone on the vertex attributes before overwriting the array with setElement. Should be a shallow clone, and we can make the loop a bit tighter. Fixed both.

- https://github.com/donmccurdy/glTF-Transform/pull/1344

## weld ✅

Use of remapPrimitive+remapAttribute resulted in iteration over the entire vertex attribute without need, even if only a subset was copied. Replaced with compactPrimitive / compactAttribute. Vastly faster on oval.gltf.

- https://github.com/donmccurdy/glTF-Transform/pull/1351
- https://github.com/donmccurdy/glTF-Transform/pull/1357

[...]

PR #1356, opened 2024-04-13
perf(functions): Merge lossy/lossless weld implementations

donmccurdy:
The newer lossless weld implementation — based on Meshoptimizer's approach — is vastly faster than the previous lossy welder, which would crash on large meshes. This PR re-implements the lossy welder using the same approach, with pre-processing to snap vertex data to a quantized grid before welding.

Remaining:

- [ ] passing tests

Related:

- #1315
- #1133


donmccurdy:
Closing in favor of https://github.com/donmccurdy/glTF-Transform/pull/1357.

Issue #1335, opened 2024-03-26
simplify: attribute weighted simplification


marwie:
Hello, currently simplify doesn't take vertex colors into account. Looks like meshoptimizer added this feature recently. Is this already exposed/available in gltf-transform?

https://github.com/zeux/meshoptimizer/issues/652#issuecomment-1922083991

donmccurdy:
Not yet exposed – see #1153. It's marked as an 'experimental' feature, so we would need to do something like that here, as I assume the meshoptimizer API could change at any time. In the meantime, vertex colors only factor in to the extent that they limit welding vertices before simplification.

marwie:
Ok - thanks for the link!

marwie:
Here's also another very very simple model where this would help (it would probably not remove any vertices in this simple case)

[BenchSmall.zip](https://github.com/donmccurdy/glTF-Transform/files/14759301/BenchSmall.zip)

![image](https://github.com/donmccurdy/glTF-Transform/assets/5083203/8f9634d8-0bc4-4752-aef9-62b2131ddbfe)

kzhsw:
Also have a look at https://github.com/donmccurdy/glTF-Transform/discussions/992

donmccurdy:
It looks like there are breaking changes planned for the simplifyWithAttributes API, which is still experimental:

https://github.com/zeux/meshoptimizer/pull/601

I think the rest of this task – supporting simplifyWithAttributes — will need to wait for a 4.1+ release.

jo-chemla:
In case it is useful to the discussion, a recent meshoptimizer [PR](https://github.com/zeux/meshoptimizer/pull/751) exposed `target_error` to `gltf_pack` https://github.com/zeux/meshoptimizer/issues/494

> This change exposes simplification error via command line argument -se, which can be valuable to control simplification more precisely and may also be used to slightly increase the defaults when -sv is used.
Also this change completes https://github.com/zeux/meshoptimizer/pull/688 by incorporating both normals and colors when -sv is specified, which improves quality on models with vertex colors further.

kzhsw:
> In case it is useful to the discussion, a recent meshoptimizer [PR](https://github.com/zeux/meshoptimizer/pull/751) exposed `target_error` to `gltf_pack` [zeux/meshoptimizer#494](https://github.com/zeux/meshoptimizer/issues/494)

[...]

PR #1357, opened 2024-04-14
chore(functions): Remove lossy weld

donmccurdy:
- BREAKING CHANGE: weld is now lossless, and does not accept tolerance options.

Avoids having to maintain two versions of weld(), the lossless version is vastly faster. Future updates to Meshoptimizer will include simplifyWithAttributes (#1335), which I hope will provide a better approach for joining vertices despite differences in vertex attributes, and allowing weld() to remain a faster, more focused, and more maintainable method.

I spent some time trying other approaches — like #1356, trying to consolidate the two methods. But that got complicated quickly. If I were trying that again, I think I'd take an approach similar to #1356 but have the QuantizedVertexStream class apply rounding when _reading_ each vertex, rather than iterating the vertex attributes up front. Then we need to rebuild the 'writeMap' after removing degenerate triangles, write a temporary index to the geometry without those triangles, then run it through compactPrimitive... it's just messier.

But I'm hoping a lossless weld is enough — that's all gltfpack appears to do for example. And when #1335 is available, that will be a (likely better) alternative for merging vertices that are not identical.

- #1315
- #1133

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -40,7 +40,6 @@ import {
 	DracoOptions,
 	simplify,
 	SIMPLIFY_DEFAULTS,
-	WELD_DEFAULTS,
 	textureCompress,
 	FlattenOptions,
 	flatten,
@@ -333,19 +332,10 @@ commands or using the scripting API.
 		validator: Validator.BOOLEAN,
 		default: true,
 	})
-	.option('--weld <bool>', 'Index geometry and merge similar vertices. Often required when simplifying geometry.', {
+	.option('--weld <bool>', 'Merge equivalent vertices. Required when simplifying geometry.', {
 		validator: Validator.BOOLEAN,
 		default: true,
 	})
-	.option(
-		'--weld-tolerance <tolerance>',
-		'Tolerance for welding vertex positions, as a fraction of primitive AABB. ' +
-			'When set to zero, welds run much faster and require bitwise equality.',
-		{
-			validator: Validator.NUMBER,
-			default: WELD_DEFAULTS.tolerance,
-		},
-	)
 	.action(async ({ args, options, logger }) => {
 		const opts = options as {
 			instance: boolean;
@@ -366,7 +356,6 @@ commands or using the scripting API.
 			flatten: boolean;
 			join: boolean;
 			weld: boolean;
-			weldTolerance: number;
 		};
 
 		// Baseline transforms.
@@ -376,15 +365,7 @@ commands or using the scripting API.
 		if (opts.palette) transforms.push(palette({ min: opts.paletteMin }));
 		if (opts.flatten) transforms.push(flatten());
 		if (opts.join) transforms.push(join());
-
-		if (opts.weld) {
-			transforms.push(
-				weld({
-					tolerance: opts.weldTolerance,
-					toleranceNormal: opts.simplify ? 0.5 : WELD_DEFAULTS.toleranceNormal,
-				}),
-			);
-		}
+		if (opts.weld) transforms.push(weld());
 
 		if (opts.simplify) {
 			transforms.push(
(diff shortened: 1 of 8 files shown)


zeux:
fwiw currently simplifyWithAttributes will still treat topological discontinuities the same as simplify; attributes are currently used to reduce attribute+geometric error during simplification but don't affect any collapse restrictions. I have some long term plans to improve this but just noting that it's not going to be a drop-in replacement to "weld with tolerance + simplify" for some time. (is it a problem for gltf-transform? not sure! gltfpack doesn't do tolerance-based welding either)

donmccurdy:
Thanks @zeux! My impression so far is that welding based on bitwise-equality before simplification works well, both here and in gltfpack, so I'm hopeful that the previous lossy welding was just unnecessary.

Another possible workaround — if anyone does want a more aggressive weld — would be to quantize before welding, like:

```javascript
import { quantize, weld, simplify } from '@gltf-transform/functions';
import { MeshoptSimplifier } from 'meshoptimizer';

await MeshoptSimplifier.ready;

await document.transform(
  quantize({
    quantizePosition: 14,
    quantizeNormal: 8,
    quantizeColor: 8
  }),
  weld(),
  simplify({
    simplifier: MeshoptSimplifier,
    error: 0.001
  })
);
```

That would provide more control than the previous tolerance-based weld, which didn't expose as much as [`QuantizeOptions`](https://gltf-transform.dev/modules/functions/interfaces/QuantizeOptions). I'll add something about that to the documentation as well.

donmccurdy:
Unfortunately, a simple counter-example of my idea is: 

1. add the basic 'monkey' mesh in blender\
2. subdivide it 5-6 times
3. export the result to GLB

Here's that mesh, compressed to meet GitHub's size restrictions:

[monkey_500k_vertices+draco.glb.zip](https://github.com/donmccurdy/glTF-Transform/files/15014852/monkey_500k_vertices%2Bdraco.glb.zip)

A lossless weld won't do anything to this model, and as a result simplification does almost nothing either – it works only where the vertices are mostly coplanar. Applying quantization first, as I suggested above, allows simplification to proceed successfully only if I reduce quantizationNormal bits to 4 (normally limited to the 8–16 range, made local changes to test). 

I suspect 4-bit normals will cause quality issues, so perhaps this quantization does need to be included as a condition in welding, rather than as a pre-processing step.

marwie:
Hello @donmccurdy I've just encountered this issue as well when running our mesh LOD generation - for testing I've been using [this model](https://engine.needle.tools/demos/gltf-progressive/assets/church-original/model.glb) 

Previously I was able to increase the threshold for lower-res LOD level which produced decent simplifcation (with some errors but that was mostly OK since those meshes would only be displayed at a distance) - see image below

image

After updating to 4.x all our LOD meshes are discarded unfortunately since weld+simplify does never exceed our simplification threshold of 5%

donmccurdy:
Probably we should try to export `quantizePrimitive` regardless of this issue. I think the API just needs a bit of cleanup first, to something like...

```javascript
import { quantizePrimitive } from '@gltf-transform/functions';

quantizePrimitive(prim, transform /* mat4 */, options /* QuantizeOptions | undefined */);
```

But, I don't think that will be a complete fix for more aggressive welding and simplification. The previous lossy weld implementation was very slow, and a completely different code path from the newer one (inspired by Meshoptimizer), so I'd prefer not to bring that back as-is. Instead I think the best approach would be to extend the VertexStream implementation here...

https://github.com/donmccurdy/glTF-Transform/blob/fc3da02a969921c341a81f10f75d7aded53725b5/packages/functions/src/hash-table.ts#L50-L59

... with a `lossyEquals` function that compares attributes based on the configured per-attribute precision, or perhaps some different precision criteria. It will be slower than the lossless mode, but hopefully not as slow as the old code.

After that, we'll need a bit of extra work to clean up degenerate triangles and remove the primitive entirely if it collapses, which can't happen in the lossless mode. I don't think `weldPrimitive` needs to handle that cleanup, but `weld()` probably should.

PR #1361, opened 2024-04-17
fixes #1360

### LINKED ISSUE #1360: Invalid bounding box calculation when one of the meshes lacks a POSITION attribute.
**Describe the bug**
When one or more of the meshes don't have the POSITION attribute, `bounds` is not calculated correctly and returns
```[ [ Infinity, Infinity, Infinity], [-Infinity, -Infinity, -Infinity] ]```
=> [getMeshBounds](https://github.com/donmccurdy/glTF-Transform/blob/main/packages/core/src/utils/get-bounds.ts#L26) returns the initial infinite bounds if the mesh has no positions. Then the [expandBounds](https://github.com/donmccurdy/glTF-Transform/blob/main/packages/core/src/utils/get-bounds.ts#L17-L18) calls here make the result an infinite bounding box.

**To Reproduce**
1. Load attached GLB
2. Get it's bounds and check

**Expected behavior**
A correct bounding box
```
min: [
    -0.45675763487815857,
    -2.7882036668004553e-8,
    -0.30479879926513953
  ],
  max: [ 
     0.4564840495586395, 
     1.8287999105357924, 
     0.28434090200747697 
  ]
```

**Versions:**
3.10.0

Some check on whether `getMeshBounds` returns a valid bounds seems in order. Or perhaps simply pass in the result bounds.

[5093-9845273-Merit 2-Drawer Wardrobe.zip](https://github.com/donmccurdy/glTF-Transform/files/15014362/5093-9845273-Merit.2-Drawer.Wardrobe.zip)

timknip:
This fixes the issue byt simply checking if min and max are valid floats.

### DIFF
--- a/packages/core/src/utils/get-bounds.ts
+++ b/packages/core/src/utils/get-bounds.ts
@@ -14,8 +14,10 @@ export function getBounds(node: Node | Scene): bbox {
 
 			// Compute mesh bounds and update result.
 			const meshBounds = getMeshBounds(mesh, node.getWorldMatrix());
-			expandBounds(meshBounds.min, resultBounds);
-			expandBounds(meshBounds.max, resultBounds);
+			if (meshBounds.min.every(isFinite) && meshBounds.max.every(isFinite)) {
+				expandBounds(meshBounds.min, resultBounds);
+				expandBounds(meshBounds.max, resultBounds);
+			}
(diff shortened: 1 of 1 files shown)


donmccurdy:
@timknip thank you!

PR #1375, opened 2024-04-27
feat(functions): Add mergeDocuments, cloneDocument, copyToDocument, moveToDocument

### LINKED ISSUE #924: Support transferring properties to another Document
The graph datastructure under glTF Transform prevents sharing property resources (materials, textures, animations...) between multiple Document instances. That's important for reference tracking, but does make certain cases harder than they need to be. For example, moving a material from one glTF file into another.

We could consider creating a simple API to handle this case. Possible APIs...

```js
const material = document.transfer(externalMaterial); // single

const materials = document.transfer(externalMaterials); // multiple

const material = externalMaterial.cloneExternal(document);
```

I think I'm leaning toward `dstProperties = document.transfer(srcProperties)`. Accepting an array would be important on the assumption that the transferred properties might share references to underlying resources like textures, and we don't want to create duplicates of those.

### LINKED ISSUE #1363: Unexpected handling of `generator` during `merge`
This might not be considered to be an "issue", but ... I stumbled over this, and find it at least unexpected: When calling `merge`, then the `asset.generator` of the _target_ document is apparently just overwritten with the one from the merged-in document.

**To Reproduce**

Run this:
```typescript
import { Document } from "@gltf-transform/core";

async function run() {
  const documentA = new Document();
  const documentB = new Document();
  const documentC = new Document();
  documentA.getRoot().getAsset().generator = "Generator A";
  documentB.getRoot().getAsset().generator = "Generator B";
  documentC.getRoot().getAsset().generator = "Generator C";
  documentA.merge(documentB);
  documentA.merge(documentC);
  console.log(documentA.getRoot().getAsset().generator);
  console.log(documentB.getRoot().getAsset().generator);
  console.log(documentC.getRoot().getAsset().generator);
}

run();
```
The output will be
```
Generator C
Generator B
Generator C
```

**Expected behavior**

That's the tough one 🙂  The `generator` is basically a "free-text" field without any specified meaning. But roughly speaking, I'd expect that ~"no information is lost". A simple and straightforward solution **could** be just just append the merged-in generator to the one that is already present, yielding 
```
Generator A, Generator B, Generator C
Generator B
Generator C
```
in the above example. **But**... with the obvious caveat that the output should't be
```
Generator X, Generator X, Generator X, Generator X, Generator X, Generator X, Generator X, Generator X
```
when merge-assembling one target document from 8 partial input documents. 

Regardless of what the solution could or will be, I thought it might be worth bringing this up...

**Versions:**
 - **3.10.1**
 - Environment: All

### LINKED ISSUE #1367: Merging documents with extensions
**Describe the bug**

This issue has two levels of abstraction: 

- How to merge two documents with `KHR_xmp_json_ld`
- How to implement extensions in a way that allows them to survive a `merge` operation in general

The second one is related to https://github.com/donmccurdy/glTF-Transform/issues/958, and _very_ specifically to [some difficulties](https://github.com/CesiumGS/3d-tiles-tools/pull/117#issuecomment-2068124358) when trying to merge documents that both contain an instance of the [`EXT_structural_metadata` implementation](https://github.com/CesiumGS/3d-tiles-tools/blob/03e5dc266efbc1f87701bcf9d124798332415538/src/gltf-extensions/gltfExtensions/EXTStructuralMetadata.ts#L243). I spent a few hours debug-stepping through various parts of the code, but no avail. 

Eventually, I started to compare the behavior to _another_ extension that defines `ROOT`-level elements - namely, `KHR_xmp_json_ld`. This leads to the first point: It looks like `KHR_xmp_json_ld` can not be merged properly...

**To Reproduce**

The following program creates two documents, each containing the `KHR_xmp_json_ld` extension, with one `packet`. It then merges these documents. The resulting document is _written_ properly. But the "root type" of the extension is `Packet` - so it's not possible to access both packets.

```typescript
import { Document } from '@gltf-transform/core';
import { NodeIO } from '@gltf-transform/core';

import { ALL_EXTENSIONS } from '@gltf-transform/extensions';
import { KHRXMP } from '@gltf-transform/extensions';
import { Packet } from '@gltf-transform/extensions';

function createTestDocument(testName: string) {
	const xmpExtension = document.createExtension(KHRXMP);
	const packet = xmpExtension.createPacket();
	packet.setContext({
		dc: 'http://purl.org/dc/elements/1.1/',
	});
	packet.setProperty('dc:Creator', { '@list': [testName] });
	const root = document.getRoot();
	root.setExtension('KHR_xmp_json_ld', packet);
	return document;
}

async function run() {
	const documentA = createTestDocument('creatorA');
	const documentB = createTestDocument('creatorB');

	document.merge(documentA);
	document.merge(documentB);

	const root = document.getRoot();
	const packet = root.getExtension<Packet>('KHR_xmp_json_ld');
	//console.log("Result:\n", result);
	const creator = packet?.getProperty('dc:Creator');
	console.log('creator', creator);

	const io = new NodeIO();
	io.registerExtensions([...ALL_EXTENSIONS]);
	const jsonDocument = await io.writeJSON(document);
	console.log(JSON.stringify(jsonDocument.json, null, 2));
}

run();
```

**Expected behavior**

The general question is: When there is an extension that defines top-level `PropertyType.ROOT` elements: Which precautions have to be taken or which "hooks" into the code can be employed to properly handle these during a `merge` operation?

For `KHR_xmp_json_ld`, one path _might_ be to have to be some (trivial) `class PacketList extends ExtensionProperty` that just is an array of `Packet` objects and serves as the top-level type for the extension. (Further details and an example are actually [given in the extension specs](https://github.com/KhronosGroup/gltf/tree/main/extensions/2.0/Khronos/KHR_xmp_json_ld#transferring-and-merging-metadata)). Then, the question will be: At which point of the `merge` process will the _contents_ of these `PacketList`-objects be merged? 

(For the more general case, I hope that any draft/sketch/idea of how this _could_ be handled for `KHR_xmp_json_ld` could help me to figure out a solution for `EXT_structural_metadata` as well 🤞 )

**Versions:**
 - Version: 3.10.1
 - Environment: All

donmccurdy:
Changes:

- BREAKING CHANGE: Removes `Document#merge` and `Document#clone` methods
- Adds `mergeDocuments(target, source)`
- Adds `cloneDocument(source)`
- Adds `copyToDocument(target, source, sourceProperties)`
- Adds `moveToDocument(target, source, sourceProperties)`

Example:

```javascript
import { copyToDocument } from '@gltf-transform/functions';

// Copy materials from sourceDocument to targetDocument.
const map = copyToDocument(
  targetDocument,
  sourceDocument,
  [sourceMaterialA, sourceMaterialB]
);

// Find counterpart of sourceMaterialA in targetDocument.
const targetMaterialA = map.get(sourceMaterialA);
```

Related:

- Fixes #924 
- Fixes #1363 
- Fixes #1367

### DIFF
--- a/benchmarks/tasks/clone.bench.ts
+++ b/benchmarks/tasks/clone.bench.ts
@@ -1,9 +1,10 @@
 import { Document } from '@gltf-transform/core';
+import { cloneDocument } from '@gltf-transform/functions';
 import { Size, Task } from '../constants';
 import { createLargeDocument } from '../utils';
 
 let _document: Document;
 
 export const tasks: Task[] = [
-	['clone', () => _document.clone(), { beforeAll: () => void (_document = createLargeDocument(Size.SM)) }],
+	['clone', () => cloneDocument(_document), { beforeAll: () => void (_document = createLargeDocument(Size.SM)) }],
 ];
--- a/packages/cli/src/transforms/merge.ts
+++ b/packages/cli/src/transforms/merge.ts
@@ -9,7 +9,7 @@ import {
 	Buffer,
 	PropertyType,
 } from '@gltf-transform/core';
-import { dedup, unpartition } from '@gltf-transform/functions';
+import { dedup, mergeDocuments, unpartition } from '@gltf-transform/functions';
 
 const NAME = 'merge';
 
@@ -41,7 +41,7 @@ const merge = (options: MergeOptions): Transform => {
 					.setMimeType(ImageUtils.extensionToMimeType(extension))
 					.setURI(basename + '.' + extension);
 			} else if (['gltf', 'glb'].includes(extension)) {
-				document.merge(renameScenes(basename, await io.read(path)));
+				mergeDocuments(document, renameScenes(basename, await io.read(path)));
 			} else {
 				throw new Error(`Unknown file extension: "${extension}".`);
 			}
--- a/packages/core/src/document.ts
+++ b/packages/core/src/document.ts
@@ -133,63 +133,24 @@ export class Document {
 		return this;
 	}
 
-	/** Clones this Document, copying all resources within it. */
+	/**
+	 * Clones this Document, copying all resources within it.
+	 * @deprecated Use 'cloneDocument(document)' from '@gltf-transform/functions'.
+	 * @hidden
+	 * @internal
+	 */
 	public clone(): Document {
-		return new Document().setLogger(this._logger).merge(this);
+		throw new Error(`Use 'cloneDocument(source)' from '@gltf-transform/functions'.`);
 	}
 
-	/** Merges the content of another Document into this one, without affecting the original. */
-	public merge(other: Document): this {
-		// 1. Attach extensions.
-		const thisExtensions: { [key: string]: Extension } = {};
-		for (const otherExtension of other.getRoot().listExtensionsUsed()) {
-			const thisExtension = this.createExtension(otherExtension.constructor as new (doc: Document) => Extension);
-			if (otherExtension.isRequired()) thisExtension.setRequired(true);
-			thisExtensions[thisExtension.extensionName] = thisExtension;
-		}
-
-		// 2. Preconfigure the Root and merge history.
-		const visited = new Set<Property>();
-		const propertyMap = new Map<Property, Property>();
-		visited.add(other._root);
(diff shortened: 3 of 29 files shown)


marwie:
Oh great, thanks a lot!

PR #1438, opened 2024-06-22
fix(functions): Remove side effects of no-op meshopt() and quantize() calls

donmccurdy:
In some cases meshopt() or quantize() may be called, but their extensions are not necessary. This may occur either because the model has no accessors, or because no primitives qualified for quantization. After this PR, extensions are not added to the document in these cases — it doesn't make sense to list KHR_mesh_quantization if there are no meshes, for example.


donmccurdy:
* **#1438** <a href="https://app.graphite.dev/github/pr/donmccurdy/glTF-Transform/1438?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #1455, opened 2024-07-11
chore(functions):  update docs in simplify

kzhsw:
Docs for calling `weld` updated, weld tolerance option is no longer available since https://github.com/donmccurdy/glTF-Transform/pull/1357


donmccurdy:
Thank you!

Issue #1454, opened 2024-07-10
platform-io: glb with unknown chunk not supported


kzhsw:
**Describe the bug**
Loading a glb with unknown chunk and without BIN chunk result in "Expected GLB BIN in second chunk." error. A quick search over code for this error leads me [here](https://github.com/donmccurdy/glTF-Transform/blob/v4.0.4/packages/core/src/io/platform-io.ts#L283).
From the [`4.4. Binary glTF Layout`](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#binary-gltf-layout) section of gltf spec, it says in [`4.4.3.1. Overview`:](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#chunks-overview)
> Client implementations **MUST** ignore chunks with unknown types to enable glTF extensions to reference additional chunks with new types following the first two chunks.

In `Table 1. Chunk types ` of spec, the `Occurrences` of BIN chunk is `0 or 1`, marking BIN chunk optional, so a model without a BIN chunk but with unknown chunk is valid from the spec.

**To Reproduce**
[test.zip](https://github.com/user-attachments/files/16154148/test.zip)
1. Go to <https://gltf.report/>
2. Load the glb in test.zip above
3. It says "Expected GLB BIN in second chunk."

**Expected behavior**
The glb should be successfully loaded and parsed.

**Versions:**
 - Version: since https://github.com/donmccurdy/glTF-Transform/commit/9d381f3794d668b6990196546788c3eec64e1f3b
 - Environment: All

**Additional context**
The Khronos [glTF 2.0 Sample Viewer](https://github.khronos.org/glTF-Sample-Viewer-Release/) can load the model without issue:
![screenshot of glTF-Sample-Viewer](https://github.com/donmccurdy/glTF-Transform/assets/120431159/cc291c32-f209-466f-8050-66672f28d372)
The [glTF-Validator](https://github.khronos.org/glTF-Validator/) result:
![glTF-Validator: The asset is valid.](https://github.com/donmccurdy/glTF-Transform/assets/120431159/d6279da7-866e-42cd-8ae9-ba6596fc3320)
Three.js [editor](https://threejs.org/editor/) can load without issue:
![screenshot of Three.js editor](https://github.com/donmccurdy/glTF-Transform/assets/120431159/98894157-326c-492c-b9bc-110de7af157f)
Babylon.js [sandbox](https://sandbox.babylonjs.com/) can load the model without issue:
[...]

PR #1457, opened 2024-07-15
fix(core): Allow GLB with unknown chunk without BIN chunk

kzhsw:
From the [`4.4. Binary glTF Layout`](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#binary-gltf-layout) section of gltf spec, it says in [`4.4.3.1. Overview`:](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#chunks-overview)
> Client implementations **MUST** ignore chunks with unknown types to enable glTF extensions to reference additional chunks with new types following the first two chunks.

In `Table 1. Chunk types ` of spec, the `Occurrences` of BIN chunk is `0 or 1`, marking BIN chunk optional, so a model without a BIN chunk but with unknown chunk is valid from the spec.

This should fix https://github.com/donmccurdy/glTF-Transform/issues/1454

### DIFF
--- a/packages/core/src/io/platform-io.ts
+++ b/packages/core/src/io/platform-io.ts
@@ -280,7 +280,9 @@ export abstract class PlatformIO {
 
 		const binChunkHeader = new Uint32Array(glb.buffer, glb.byteOffset + binByteOffset, 2);
 		if (binChunkHeader[1] !== ChunkType.BIN) {
-			throw new Error('Expected GLB BIN in second chunk.');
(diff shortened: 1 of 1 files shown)


donmccurdy:
Thank you!

donmccurdy:
Added a small test to cover this case as well:

- https://github.com/donmccurdy/glTF-Transform/commit/323c859441e5e9eef7ad20542d74d74156c18b1c

PR #1491, opened 2024-08-26
fix(view/lights): for spotlights, restore parent on its target

### LINKED ISSUE #1490: Updating spotlight's definition, detaches the spotlight target.
**Describe the bug**
In `@glTF-transform/view`, when updating the light's definition, causes the target to no longer update, with parent's transformation.

**To Reproduce**
1. Create a light
2. Attach to a node.
3. Update any property on light.
4. Target no longer responds to updates to the node, like rotation.

```
const lightsExt = doc.createExtension(KHRLightsPunctual);
const light = lightsExt.createLight('Light 1');
const node = doc.createNode('Node 1');
node.setExtension("KHR_lights_punctual", light);

let lightObj = docView.view(light);
lightObj.target.parent; // correct

light.setIntensity(10);

let lightObj = docView.view(light);
lightObj.target.parent; // undfined
```

**Expected behavior**
Transformations to the parent should continue to reflect on the target.

**Versions:**
 - Version: 4.0.8
 - Environment: Browser

divyun:
Resolves #1490 

Target is added as a child to light. https://github.com/donmccurdy/glTF-Transform/blob/23e2bc6c2aae0dd7cdbb1cf589bfa125b6768426/packages/view/src/subjects/LightSubject.ts#L50
But when the lights are cloned here
https://github.com/donmccurdy/glTF-Transform/blob/23e2bc6c2aae0dd7cdbb1cf589bfa125b6768426/packages/view/src/pools/SingleUserPool.ts#L29, the corresponding clone in three.js https://github.com/mrdoob/three.js/blob/817a222f2d12baf44a38baf256bc9d4f65d82465/src/lights/SpotLight.js#L61 doesn't add a parent to target.

### DIFF
--- a/packages/view/src/pools/SingleUserPool.ts
+++ b/packages/view/src/pools/SingleUserPool.ts
@@ -1,5 +1,5 @@
 import { type Mesh as MeshDef, type Node as NodeDef, type Property as PropertyDef, uuid } from '@gltf-transform/core';
-import type { Object3D } from 'three';
+import { DirectionalLight, type Object3D, SpotLight } from 'three';
 import type { LightLike } from '../constants.js';
 import { Pool } from './Pool.js';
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
Thanks @divyun, and sorry for the delay on this!

donmccurdy:
Minor followup in https://github.com/donmccurdy/glTF-Transform/pull/1699.

Issue #1509, opened 2024-09-24
Object transforms look off after using Instance


Kyozon:
**Describe the bug**
The resulting .gltf scene looks off when using the `gltf-transform instance`.

**To Reproduce**
1. Download: [merged.zip](https://github.com/user-attachments/files/17120741/merged.zip)

2. Import `merged.gltf` into Blender and verify original transforms as seen in the image below:

![Default](https://github.com/user-attachments/assets/7102a7a8-3cf4-4b18-8bb9-a6d594602369)

3. Use `gltf-transform instance merged.gltf instanced.gltf`

4. Import `instanced.gltf` into Blender and verify the bug as seen in the image below:

![Instance](https://github.com/user-attachments/assets/f6c8d85a-70bc-422a-a9cd-104f40acb332)

**Expected behavior**
Retain the original transforms but optimize the scene to use GPU instances. 

**Versions:**
 - Version: 4.0.8
 - Environment: Node.js CLI

donmccurdy:
Hi @Kyozon! The instanced scene looks correct to me, both in Blender and in https://gltf-viewer.donmccurdy.com/. Could you confirm if you're seeing the problem both in Blender and in my viewer, and perhaps share the 'broken' copy of the model as well? 

![CleanShot 2024-09-24 at 16 40 19@2x](https://github.com/user-attachments/assets/459089a5-cded-435c-87b5-dcfd05102789)

Kyozon:
Hi @donmccurdy. Thank you for the quick reply! 

Sure, here is the broken file: [Instanced.zip](https://github.com/user-attachments/files/17121000/Instanced.zip)

donmccurdy:
Hm... something is weird in the source file:

```jsonc
  "buffers": [
    {
      "uri": "glTF.bin",
      "byteLength": 8160
    },
    {
      "uri": "glTF.bin",
      "byteLength": 8160
    },
    {
      "uri": "glTF.bin",
      "byteLength": 8160
    }
  ],
```

Three buffers, all pointing at the same thing... I'm not sure whether it's technically valid in the glTF spec, but the pygltflib authors probably didn't intend that, and it's leading to data loss here. 

I'll have to think a bit about the best way to handle it, but in the meantime, you can avoid the issue by exporting to `.glb` instead of `.gltf`.

donmccurdy:
[...]

PR #1511, opened 2024-09-28
fix(io): Handle duplicate URIs on read/write

### LINKED ISSUE #1510: Normal ranges are modified after processing
**Describe the bug**
Transforming (eg "gltf-transform flatten") a glb with [ - 1, 1 ] range float32 normals results in out-of-range values.

I'm seeing _massively_ out of range values on my internal test models - going from [ - 1, 1 ] to [ - 440, 4505] or larger - but I can't share those here. However using a simpler model exported from the three.js editor you can see that the normal data seems to be mangled in some way.

Here's a sample model which are these three objects inside a large sphere. I'm not exactly sure what's causing the issues so it's hard to repro:

[repro.glb.zip](https://github.com/user-attachments/files/17126997/repro.glb.zip)

image

**To Reproduce**
1. Drag the "repro.glb" into the glb viewer and log the "normal" attribute ranges. See they are in the range [ - 1, 1 ] and [ 0, 1 ]
2. Run "gltf-transform flatten" on the repro.glb file
4. Log the normal ranges again
5. See the ranges are [ - 1, 1.015625 ], [ - 1, 1 ]

**Expected behavior**
The range of normals is not changed.

**Versions:**
 - Version: v4.0.8
 - Environment: node js (cli)

**Additional context**

donmccurdy:
- Fixes https://github.com/donmccurdy/glTF-Transform/issues/1509
- Related https://github.com/KhronosGroup/glTF/issues/2446

Changes:

- While reading, distinct resources (buffers, images) with the same URIs remain distinct, but the URI is removed from all but the first, such that a new URI will be generated for the remainder when writing. A warning is logged.
- While writing, a warning is logged if two resources share the same URI. If the two resources are not identical, an error is thrown.

To explicitly consolidate duplicate resources, the [`dedup()`](https://gltf-transform.dev/modules/functions/functions/dedup) function may be used.

### DIFF
--- a/packages/core/src/io/reader-context.ts
+++ b/packages/core/src/io/reader-context.ts
@@ -1,4 +1,5 @@
 import type { JSONDocument } from '../json-document.js';
+import type { Document } from '../document.js';
 import type {
 	Accessor,
 	Animation,
@@ -13,6 +14,7 @@ import type {
 	TextureInfo,
 } from '../properties/index.js';
 import type { GLTF } from '../types/gltf.js';
+import type { ILogger } from '../utils/logger.js';
 
 /**
  * Model class providing glTF Transform objects representing each definition in the glTF file, used
@@ -22,6 +24,11 @@ import type { GLTF } from '../types/gltf.js';
  * @hidden
  */
 export class ReaderContext {
+	public readonly document: Document;
+	// TODO(v5): Rename to jsonDocument;
+	public readonly jsonDoc: JSONDocument;
+	private readonly logger: ILogger;
+
 	public buffers: Buffer[] = [];
 	public bufferViews: Uint8Array[] = [];
 	public bufferViewBuffers: Buffer[] = [];
@@ -36,7 +43,13 @@ export class ReaderContext {
 	public animations: Animation[] = [];
 	public scenes: Scene[] = [];
 
-	constructor(public readonly jsonDoc: JSONDocument) {}
+	private _usedURIs = new Set<string>();
+
+	constructor(document: Document, jsonDoc: JSONDocument) {
+		this.document = document;
+		this.jsonDoc = jsonDoc;
+		this.logger = document.getLogger();
+	}
 
 	public setTextureInfo(textureInfo: TextureInfo, textureInfoDef: GLTF.ITextureInfo): void {
 		this.textureInfos.set(textureInfo, textureInfoDef);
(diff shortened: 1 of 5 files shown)


javagl:
A related question came up for me in JglTF recently. In that context, I did a few experiments with glTF-Transform, and referred to that in a comment at https://github.com/javagl/JglTF/issues/115#issuecomment-2197347281 

The test that I did there, and where one could argue about whether it is a "bug" or only a "maybe somewhat surprising behavior" of glTF-Transform is this:
```
import fs from "fs";
import { Document } from "@gltf-transform/core";
import { NodeIO } from "@gltf-transform/core";

async function run() {
  const io = new NodeIO();
  document.createBuffer();

  const image0 = fs.readFileSync("image0.png");
  const texture0 = document.createTexture();
  texture0.setMimeType("image/png");
  texture0.setImage(image0);

  const image1 = fs.readFileSync("image1.png");
  const texture1 = document.createTexture();
  texture1.setMimeType("image/png");
  texture1.setImage(image1);

  //texture0.setURI("test.png");
  texture1.setURI("texture_1.png");

  const output = await io.writeJSON(document);
  console.log(JSON.stringify(output.json, null, 2));
  console.log(output.resources);
}

run();
```

It **only** sets the URI of the second texture, via `texture1.setURI("texture_1.png");`, and this causes the _first_ texture to be overwritten, because it happens to receive the auto-generated URI `"texture_1.png"` as well....

---

Crying wold: The part in the comment above that says

> ... the URI is removed from all but the first,

sounds like it could cause unexpected behavior, but I haven't looked into the details yet.

donmccurdy:
Reverting in https://github.com/donmccurdy/glTF-Transform/commit/84e77f58d321702dba486fc07ed1ce7279d4897d, due to #1513. 

The immediate the problem seems to be the "import" side, though I think you raise some good points that the export side might be better served if unique URL generation were aware of every other resource URL in the document, and not just the unnamed resources.

donmccurdy:
Updates to this PR coming in:

- https://github.com/donmccurdy/glTF-Transform/pull/1522

I may consider making this handling more strict in v5 or another major release, whether it's considered invalid by the spec or not, but still TBD.


> It only sets the URI of the second texture, via texture1.setURI("texture_1.png");, and this causes the first texture to be overwritten, because it happens to receive the auto-generated URI "texture_1.png" as well....

Currently happening here:

https://github.com/donmccurdy/glTF-Transform/blob/93ca869a865712baf4d615f7393d053a480de0f3/packages/core/src/io/writer-context.ts#L233-L252

The result won't always be `texture_1`, it would be `<slot>_1.png` if the texture were assigned to a material, but still conflicts are possible if the input scene has a mix of external and embedded (no URI) textures, or if textures are added through the scripting API without assigning unique URIs. Improvements here are welcome but not currently planned.

PR #1526, opened 2024-10-11
feat(extensions,functions): Allow encoding meshopt-compressed float33 accessors

### LINKED ISSUE #1524: Allow writing EXT_meshopt_compression without quantization
While meshopt compression (EXT_meshopt_compression) is generally used alongside quantization (KHR_mesh_quantization) for best compression and memory optimization, there's no technical requirement that they be used together. Currently glTF Transform just doesn't support that, it always does quantization when doing meshopt compression.

This issue tracks a feature request for allowing meshopt compression without quantization. Compressed sizes will be larger, but compatibility and quality will be somewhat higher.

donmccurdy:
- fixes #1523 
- fixes #1524

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -254,7 +254,7 @@ commands or using the scripting API.
 		default: INSTANCE_DEFAULTS.min,
 	})
 	.option('--meshopt-level <level>', 'Meshopt compression level.', {
-		validator: ['medium', 'high'],
+		validator: ['low', 'medium', 'high'],
 		default: 'high',
 	})
 	.option('--palette <bool>', 'Creates palette textures and merges materials.', {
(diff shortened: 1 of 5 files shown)


zeux:
FWIW while two out of three filters that EXT_meshopt_compression provides are designed for quantized data (normals & quaternions), the exponential filter is designed for float32 data and can be used without KHR_mesh_quantization; gltfpack will do this when asked (using `-vpf` or `-vtf` flags for positions or texture coordinates respectively). Only noting this because it looks like this PR assumes filters must be applied to quantized accessors.

The value of using exponential filter is that it is transparent to the application (as it's just a transform that the rest of glTF document doesn't observe) but it allows to smoothly adjust compression ratio with some precision loss (controlled via bits & mode). The compression ratio (and VRAM savings!) are still the best with quantized data, so it's not a full replacement, but a useful alternative.

donmccurdy:
@zeux Really helpful, thank you! I see I have some incorrect assumptions in this PR — I may need to rethink the API currently exposed for configuring EXT_meshopt_compression.

Issue #1530, opened 2024-10-14
Transform entire scene


rotu:
**Is your feature request related to a problem? Please describe.**
I have a bunch of models which may be rotated and scaled incorrectly, and gltf-transform does not provide an easy way to fix this.

**Describe the solution you'd like**
I would like a new file function that inserts a new root node in all scenes with a given transform.

e.g. `gltf-transform reroot --translate [0,0,1] --rotate [1,0,0,1] --scale 0.4`
(where the argument to rotate is a non-normalized quaternion)

**Describe alternatives you've considered**
I currently do this with a script like following:

```js
document.getRoot().listScenes().forEach(
  s=>{
    s.listChildren().forEach(c=>{
    const newRoot = document.createNode();
    newRoot.setTranslation([0,0,1])
    newRoot.setRotation(normalize([1,0,0,1]));
    newRoot.setScale([0.4,0.4,0.4])
    newRoot.addChild(c);
    s.addChild(newRoot)
  })}
)

function normalize(v){
  const n = 1/Math.hypot(...v)
  if (!Number.isFinite(n)){ throw new Error() }
  return Array.from(v, x=>x*n)
}
```

The `translate` feature has much in common with the `center` function, so it may make sense to unify these.

rotu:
Here's a cleaned up and parameterized implementation:

```ts
import { Document, Transform, vec3, vec4 } from "@gltf-transform/core";
import { assignDefaults, createTransform } from "@gltf-transform/functions";

const NAME = "transformWorld";

type TransformWorldOptions = {
  translation?: vec3;
  rotation?: vec4;
  scale?: number | vec3;
};
const TRANSFORM_WORLD_DEFAULTS: Required<TransformWorldOptions> = {
  translation: [0, 0, 0] as vec3,
  rotation: [0, 0, 0, 1] as vec4,
  scale: 1,
};

function normalize(v: vec4): vec4;
function normalize(v: number[]) {
  const n = 1 / Math.hypot(...v);
  return Array.from(v, (x) => x * n);
}
function uniform(s): vec3 {
  return [s, s, s];
}
function transformWorld(
  _options: TransformWorldOptions = TRANSFORM_WORLD_DEFAULTS,
): Transform {
  const options = assignDefaults(TRANSFORM_WORLD_DEFAULTS, _options);
  return createTransform(NAME, (doc: Document): void => {
    const logger = doc.getLogger();
    doc
      .getRoot()
      .listScenes()
[...]

PR #1536, opened 2024-10-18
Add spliceNode function to reparent an entire scene or subtree

rotu:
This is inspired by https://github.com/donmccurdy/glTF-Transform/issues/1530#issuecomment-2414064338. In particular, instead of a `transformScene` function, this creates a new `Node` to which you can apply transforms.

### DIFF
--- a/packages/functions/src/splice-node.ts
+++ b/packages/functions/src/splice-node.ts
@@ -1,0 +1,28 @@
+import { Scene, Node } from '@gltf-transform/core';
+
+/**
+ * Create a new node as child of the target {@link Scene} or {@link Node}.
+ * Its existing children will be reparented under the new node.
+ *
+ * Example:
+ *
+ * ```javascript
+ * import { transformScene } from '@gltf-transform/functions';
+ *
+ * let scene = document.getRoot().getDefaultScene()
+ * // rotate the entire scene a quarter turn around the y axis
+ * spliceNode(scene).setRotation([0, Math.SQRT1_2, 0, Math.SQRT1_2]);
+ * ```
+ *
+ * @param parent Node or Scene whose children to reroot
+ * @returns The new parent node
+ */
+export function spliceNode(parent: Scene | Node): Node {
+	const newNode = new Node(parent.getGraph());
+	for (const child of parent.listChildren()) {
+		if (child.getSkin()) continue;
+		newNode.addChild(child);
+	}
+	parent.addChild(newNode);
+	return newNode;
+}


rotu:
I'm not sure about the API. In particular, the `const newNode = new Node(parent.getGraph())`. I'm not sure if it's kosher for a function to create a `Node` object outside of `document.createNode()`. The alternative is to require the caller to pass in the new parent, e.g. 
```
const newParent = document.createNode().setRotation([1,0,0,0]);
spliceNode(oldParent, newParent);
```
but then it's not clear what to do if `newParent` already *has* a parent. Or worse, if `newParent` is already a child of `oldParent`.

Also not totally happy with the name of this function.

donmccurdy:
> I'm not sure about the API. In particular, the const newNode = new Node(parent.getGraph()).

It's OK for the function to create a node, I think, as long as the naming is consistent with that expectation. Calling the `Node` constructor (or other Property subclass constructors) directly will cause issues though, I would copy the pattern here:

https://github.com/donmccurdy/glTF-Transform/blob/1693ffdb8245c1765ab019d566f58dffb6a56093/packages/functions/src/list-texture-slots.ts#L13-L16

I'd like to make that more obvious at some point, but in the meantime this avoids requiring users to pass in a `document` to functions unnecessarily.

That said, having the user pass in both nodes would sound OK to me too.

packages/functions/src/splice-node.ts
```diff
@@ -0,0 +1,28 @@
...
+ * spliceNode(scene).setRotation([0, Math.SQRT1_2, 0, Math.SQRT1_2]);
+ * ```
+ *
+ * @param parent Node or Scene whose children to reroot
+ * @returns The new parent node
+ */
+export function spliceNode(parent: Scene | Node): Node {
+	const newNode = new Node(parent.getGraph());
+	for (const child of parent.listChildren()) {
+		if (child.getSkin()) continue;
```
donmccurdy, line 23: Thinking more about my recommendation from https://github.com/donmccurdy/glTF-Transform/issues/1530#issuecomment-2414064338 here, and unfortunately it's not as simple as I'd been hoping. 

1. As mentioned, we will get validation warnings if we put a parent above a skinned mesh. Technically it's valid, these are just 'hints' or 'warnings' at the validator level, because the transform won't do anything to the skinned mesh _directly_, transforms affect only the joints.
2. However, it's possible that the skinned mesh's joints are descendants of the node. In that case, skipping the skinned mesh will cause the transform to not affect the joints as expected, and thus not to affect the skinned mesh.

Probably (2) is the more critical error, as it will change the rendered result.

We could search the child's subtree for any joints, and skip the skinned mesh only if no joints are found. But this might be a more complex and less predictable operation than users are expecting, calling a method like 'spliceNode'. If we skip that special case, then the method body becomes something more like:

```javascript
const node = Document.fromGraph(parent.getGraph()).createNode();
for (const child of parent.listChildren()) {
  node.addChild(child);
}
parent.addChild(newNode);
```

rotu:
> It's OK for the function to create a node, I think, as long as the naming is consistent with that expectation. Calling the `Node` constructor (or other Property subclass constructors) directly will cause issues though, I would copy the pattern here:
> 
> https://github.com/donmccurdy/glTF-Transform/blob/1693ffdb8245c1765ab019d566f58dffb6a56093/packages/functions/src/list-texture-slots.ts#L13-L16
> 
> I'd like to make that more obvious at some point, but in the meantime this avoids requiring users to pass in a `document` to functions unnecessarily.

Here are some passing notes about how I was confused. I'm sure more hacking on this will clear things up:

- I didn't use `fromGraph` because I inferred from the name (incorrectly, but in analog with `Array.from(...)`) that it creates a *new* `Document`. It also has the problem that it can return `null` and it's not clear what to do if the node is merely part of a document fragment.
- `getGraph` is fuzzy about whether the returned object is necessarily the same for all `GraphNode`s, or if it's just the reachable subgraph (subtree for `Node`, since the node hierarchy is acyclic)
- I'm super uncertain about the distinction between a `Graph` and a `Document` and what's needed in order to maintain class invariants. I called the constructor that way because that's what the [`createNode` function seems to do](https://github.com/donmccurdy/glTF-Transform/blob/1693ffdb8245c1765ab019d566f58dffb6a56093/packages/core/src/document.ts#L205-L207).

PR #1628, opened 2025-03-21
Add `predicate` option to filter meshes in `join` function.

mysticfall:
The `predicate` function allows users to define a condition to filter meshes during the processing. It defaults to a function that always returns `true`, ensuring backward compatibility while providing more customization options.

### DIFF
--- a/packages/functions/src/join.ts
+++ b/packages/functions/src/join.ts
@@ -53,12 +53,22 @@ export interface JoinOptions {
 	 * @experimental
 	 */
 	cleanup?: boolean;
+	/**
+	 * A filter function used to evaluate a condition on a given {@link Node Node}.
+	 * This function should return a boolean indicating whether the node
+	 * satisfies the provided condition.
+	 *
+	 * @param {Node} node - The node instance to be evaluated.
+	 * @returns {boolean} - The result of the evaluation; `true` if the condition is met, otherwise `false`.
+	 */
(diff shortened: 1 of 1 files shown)


mysticfall:
Probably, it would be better to add the same option to all similar functions. But for now, I just need it for `join`, which is necessary for me to use glTF-Transform for my work project.

donmccurdy:
Hi @mysticfall, thanks for this PR! 

I'd be happy to merge this, but have two bits of feedback:

- I might prefer to name the parameter `filter`, to me it sounds more common
- For your use cases, would a condition on the Node work just as well? I think that might be more flexible in case you need to check the node name or extras, and the mesh would still be accessible in the filter function as `node.getMesh()`

mysticfall:
> Hi @mysticfall, thanks for this PR!
> 
> I'd be happy to merge this, but have two bits of feedback:
> 
>     * I might prefer to name the parameter `filter`, to me it sounds more common
> 
>     * For your use cases, would a condition on the Node work just as well? I think that might be more flexible in case you need to check the node name or extras, and the mesh would still be accessible in the filter function as `node.getMesh()`

Thanks for the review. I agree with you on both points and just updated my PR accordingly.

Let me know if there's anything more I should change. Thanks!

donmccurdy (approved):

donmccurdy:
Thank you @mysticfall!

PR #1681, opened 2025-07-06
feat(functions): Allow instance even if some nodes are animated

### LINKED ISSUE #1626: Support instance() operations on partially-animated scenes
**Describe the bug**
`gltf-transform instance` refuses to apply `EXT_mesh_gpu_instancing` instancing, if even one object in the scene has an animation.

Here I have prepared some test files for reproducing this: 

[25Spheres.zip](https://github.com/user-attachments/files/19281715/25Spheres.zip)

Here I have a Blender `4.3` file with 25 spheres. `25Sphere.blend` in the zip file. I export it as a .glb scene. `25Spheres.glb` in the zip file. No instancing yet, as reported by ` gltf-transform inspect .\25Spheres.glb`.

Now I apply `gltf-transform instance .\25Spheres.glb .\25SpheresInstanced.glb`, again attached in the .zip file. Great, now it has instancing.
```
gltf-transform inspect .\25SpheresInstanced.glb                                                                                                                                                                                                         
 OVERVIEW
 ────────────────────────────────────────────
┌────────────────────┬─────────────────────────┐
│ key                │ value                   │
├────────────────────┼─────────────────────────┤
│ version            │ 2.0                     │
├────────────────────┼─────────────────────────┤
│ generator          │ glTF-Transform v4.1.2   │
├────────────────────┼─────────────────────────┤
│ extensionsUsed     │ EXT_mesh_gpu_instancing │
├────────────────────┼─────────────────────────┤
│ extensionsRequired │ none                    │
└────────────────────┴─────────────────────────┘
```

However, now I add an completely unrelated animated cube: `25Spheres-With-AnimatedCube.blend` in the .zip file. And I export the scene to a .glb file `25Spheres-With-AnimatedCube.glb` in the zip file. Now i run the same `gltf-transform instance .\25Spheres-With-AnimatedCube.glb .\25Spheres-With-AnimatedCube-Instanced.glb` Output in the zip file.

Now there is a warning: `warn: instance: Instancing is not currently supported for animated models.` Inspect reports:
```
gltf-transform inspect .\25Spheres-With-AnimatedCube-Instanced.glb                                                                                                                                                                                      
 OVERVIEW
 ────────────────────────────────────────────
┌────────────────────┬───────────────────────┐
│ key                │ value                 │
├────────────────────┼───────────────────────┤
│ version            │ 2.0                   │
├────────────────────┼───────────────────────┤
│ generator          │ glTF-Transform v4.1.2 │
├────────────────────┼───────────────────────┤
│ extensionsUsed     │ none                  │
├────────────────────┼───────────────────────┤
│ extensionsRequired │ none                  │
└────────────────────┴───────────────────────┘
```

Instancing can be applied to the 25 spheres as previously. The animated model is unrelated to this, yet `gltf-transform` refuses based on the fact that the scene has at least one animated model. I would consider this a bug, as I presume instancing to work in a scene, on un-animated models, even if there are unrelated animated models present.

**To Reproduce**
Use the attached files to reproduce as described above

**Expected behavior**
`instance` should apply instancing, even if there exists at least one unrelated object with animations.

**Versions:**
 - Version: GLTF-Transform: `4.1.2`
 - Environment: Conversion stage, so inapplicable

**Additional context**

matthewhilton:
Partially fixes #1626 

Allows instancing to run even if some of the nodes in the scene are animated.

For example: In my project, the camera is animated but everything else is static, so having everything else able to be instanced is a huge performance win.

I lack the knowledge on the gltf spec to know if this covers every edge case though.

### DIFF
--- a/packages/functions/src/instance.ts
+++ b/packages/functions/src/instance.ts
@@ -49,11 +49,13 @@ export function instance(_options: InstanceOptions = INSTANCE_DEFAULTS): Transfo
 		const logger = doc.getLogger();
 		const root = doc.getRoot();
 
-		if (root.listAnimations().length) {
-			logger.warn(`${NAME}: Instancing is not currently supported for animated models.`);
-			logger.debug(`${NAME}: Complete.`);
-			return;
-		}
+		// Animated nodes are not supported yet for instancing - so find which ones are animated.
+		const allAnimationTargetNodes = root
+			.listAnimations()
+			.map(a => a.listChannels())
+			.flat()
+			.map(c => c.getTargetNode())
+			.filter(n => n != null)
 
 		const batchExtension = doc.createExtension(EXTMeshGPUInstancing);
 
@@ -64,6 +66,7 @@ export function instance(_options: InstanceOptions = INSTANCE_DEFAULTS): Transfo
 			// Gather a one-to-many Mesh/Node mapping, identifying what we can instance.
 			const meshInstances = new Map<Mesh, Set<Node>>();
 			scene.traverse((node) => {
+				if (allAnimationTargetNodes.includes(node)) return;
 				const mesh = node.getMesh();
 				if (!mesh) return;
 				if (node.getExtension('EXT_mesh_gpu_instancing')) return;
--- a/packages/functions/test/instance.test.ts
+++ b/packages/functions/test/instance.test.ts
@@ -1,4 +1,4 @@
-import { Document } from '@gltf-transform/core';
+import { Accessor, AnimationChannel, Document } from '@gltf-transform/core';
 import { EXTMeshGPUInstancing, type InstancedMesh } from '@gltf-transform/extensions';
 import { instance } from '@gltf-transform/functions';
 import { createTorusKnotPrimitive, logger } from '@gltf-transform/test-utils';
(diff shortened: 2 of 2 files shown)


packages/functions/src/instance.ts
```diff
@@ -64,6 +66,7 @@ export function instance(_options: InstanceOptions = INSTANCE_DEFAULTS): Transfo
 			// Gather a one-to-many Mesh/Node mapping, identifying what we can instance.
 			const meshInstances = new Map<Mesh, Set<Node>>();
 			scene.traverse((node) => {
+				if (allAnimationTargetNodes.includes(node)) return;
```
donmccurdy, line 69: Optional, a simpler check might be looking for references to the node from animations:

```suggestion
				const isAnimated = node.listParents().some((p) => p instanceof AnimationChannel);
				if (isAnimated) return;
```

But this might conflict with my other comment, so feel free to ignore. :)

donmccurdy (commented):
Thanks @matthewhilton! I think one other case that would be important to handle is when a node has animated ancestors in the scene graph. In that case we need to skip the entire subtree of children under the animated node. It might be worthwhile to make a small inline copy of the `traverse` function, allowing the callback to return a boolean true/false indicating whether to continue traversing that node's children:

https://github.com/donmccurdy/glTF-Transform/blob/9bb441be72dc824917deafb00f8f1a06d00d14db/packages/core/src/properties/node.ts#L287-L291

Issue #1678, opened 2025-07-02
CLI Optimize: --compress draco flag overrides --weld false


Eritar:
**Describe the bug**
CLI Optimize: --compress draco flag overrides --weld false

**To Reproduce**
1. Use the command `gltf-transform optimize 1.glb 2.glb --weld false --compress draco -v`
2. Use the command `gltf-transform optimize 1.glb 2.glb --weld false --compress false -v` for comparison.

**Expected behavior**
Welding operations to behave independently from draco compression.

**Versions:**
 - Version: 4.2.0
 - Environment: Windows 11, cmd cli

**Additional context**
Operation results:
[wo_draco.txt](https://github.com/user-attachments/files/21018310/wo_draco.txt)
[w_draco.txt](https://github.com/user-attachments/files/21018311/w_draco.txt)

Thank you for your work, and for your amazing software!

donmccurdy:
Hi @Eritar! It's my understanding that Draco compression requires indexed primitives, so welding is required for Draco compression.

donmccurdy:
Alternatively, meshopt compression should behave more as you're probably expecting. To do an apples-to-apples size comparison with Draco, you'd want to gzip the entire .glb after applying meshopt.

PR #1688, opened 2025-07-16
option to not weld - just do draco

zrshannes:
image
We have this scenario where we don't want to remove unused stuff, but still want the mesh to be compressed with Draco.
So, we need Draco to not weld at the start of a Draco compression.

### DIFF
--- a/packages/functions/src/draco.ts
+++ b/packages/functions/src/draco.ts
@@ -6,6 +6,7 @@ import { weld } from './weld.js';
 const NAME = 'draco';
 
 export interface DracoOptions {
+	skipWeld?: boolean;
 	method?: 'edgebreaker' | 'sequential';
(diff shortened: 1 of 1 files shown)


donmccurdy:
Hi @zrshannes! For context, welding (or at least indexing) is required for Draco compression:

- https://github.com/donmccurdy/glTF-Transform/issues/1678#issuecomment-3028501110

If your concern is really with the cleanup applied during weld() — pruning accessors and nodes — and not with the weld operation itself, then that should be addressed soon in:

- https://github.com/donmccurdy/glTF-Transform/pull/1685

Or, if I've missed something about the issue you're facing, please let me know. :)

donmccurdy:
Closing in favor of #1685 unless we find anything necessary is missing, thanks!

Issue #1689, opened 2025-07-16
View > Emissive Strength doesn't get applied to corresponding three.js material


divyun:
**Describe the bug**
When loading a glTF with `KHR_materials_emissive_strength` extension, emissive intensity on the corresponding three.js material stays `1`, regardless of the value of `emissiveStrength` defined in the extension.

**To Reproduce**
1. Load a glTF with a material that has  `KHR_materials_emissive_strength` extension enabled and `emissiveStrength` set to 3.
2. Request its three.js material using `documentView.view(mat)`.
3. Value of the `emissiveIntensity` for the three.js material should be 3 but is 1 instead.

Here is a reproduction of the bug: https://jsfiddle.net/ynvx51mr/2/

**Expected behavior**
Value of the `emissiveIntensity` for the three.js material should be equal to the `emissiveStrength` set on the extension

**Versions:**
 - Version: 4.2
 - Environment: Browser

PR #1690, opened 2025-07-16
🐛(view/material): fix emissiveIntensity when requesting material's threejs obj

divyun:
Fixes https://github.com/donmccurdy/glTF-Transform/issues/1689

In case of `KHR_materials_emissive_strength`, `getShadingModel` returns to be a standard material but currently, `emissiveIntensity` doesn't update unless its a physical material. I've updated it to compute `emissiveIntensity` for a standard material.


donmccurdy:
Thanks @divyun! I've added emissive strength to the examples ...

- https://github.com/donmccurdy/glTF-Transform/pull/1694 

... and this is working well!

Issue #1250, opened 2024-01-27
Create a microblog


donmccurdy:
I'm interested in creating a microblog for glTF Transform, with short snippets and examples. This could go on the website, perhaps with an RSS feed. Aiming for stuff that's lower effort than my own personal blog posts. Related issues:

- #984 

Ideas / feedback / upvotes / downvotes all welcome.

Issue #984, opened 2023-06-12
Enhanced Tutorials


donmccurdy:
This thread is meant to gather and organize ideas about new examples and tutorials to be added to the glTF Transform documentation (https://gltf-transform.donmccurdy.com/). Suggests, comments, questions, and upvotes are welcome. Requests for new feature development should probably be made in a new issue, instead.

Ideas:

- [ ] Building custom extensions
- [ ] Building custom pipelines
- [ ] Baking ambient occlusion with [three-gpu-pathtracer](https://github.com/gkjohnson/three-gpu-pathtracer/)
- [ ] Integrating drag-and-drop input to glTF Transform in a website
- [ ] Processing 3D Tiles

dasmehdix:
I have generally use `gltf-transform/CLI` tool. I am not much familiar with Javascript. When I want to use `gltf-transform inspect`, I would need to get a Json file that contains properties of glb/gltf model. Reading console output can be a bit tough. So, can you provide an argument that save inspect output as Json or Csv?

donmccurdy:
Hi @dasmehdix – Do you mind filing a new issue for the feature request? I'm hoping to keep this thread focused on adding tutorials to the site documentation. What you can also do now would be...

```
gltf-transform inspect input.glb --format csv
```

... which prints CSV-formatted content to the CLI, and you can paste that into Excel or Google Sheets. Or `--format md` will get you markdown tables. It isn't possible to write the entire report to a single CSV, as the different tables have different structures, and I didn't feel great about writing CSVs in bulk for this command.

jo-chemla:
Hi there, interested to know what tutorial you would like to write regarding `Processing 3d-tiles` - from my understanding the glTF-Transform utility was focused on IO/transforms/optimization tooling on glTF rather than a hierarchy of glTF tiles. Do you plan to describe how to merge glb/glTF tiles from a single 3D-tiles tileset? And a warm thanks for your great tool!

donmccurdy:
@jo-chemla totally undecided at this point! I am hoping for feedback, and may not write anything at all without it. While I don't anticipate the glTF Transform library adding new features specifically for the 3D Tiles format,  glTF Transform is an I/O library and could be used to write converters to/from other formats. Combined with another library or custom code that reads/writes 3D tiles, for example, you could presumably convert between the two. An in-depth tutorial explaining how such a pipeline would be built is what I'm mulling over here.

javagl:
[...]

PR #1719, opened 2025-09-19
feat: instance method supports custom properties

WallanceLee:


donmccurdy:
Hi @WallanceLee! Thanks for the PR. I'm sorry for any confusion, as my comment in https://github.com/donmccurdy/glTF-Transform/discussions/1715#discussioncomment-14456954 might not have been clear. I think your implementation is good, but unfortunately I'm not prepared to maintain this new feature within the glTF Transform library right now.

I'll close the PR, but I do think your approach of customizing the `instance()` function to your needs is right, and this could be a good candidate for a tutorial in the future, see:

- https://github.com/donmccurdy/glTF-Transform/issues/984
- https://github.com/donmccurdy/glTF-Transform/issues/1250

PR #1720, opened 2025-09-22
feat: primive-outline

WallanceLee:


packages/functions/src/primitive-outline.ts
```diff
@@ -0,0 +1,135 @@
...
+};
+
+export function createPrimitiveOutline(_option: PrimitiveOutlineOptions = PRIMITIVE_OUTLINE_DEFAULTS): Primitive {
+	const options = assignDefaults(PRIMITIVE_OUTLINE_DEFAULTS, _option);
+
+	return createTransform(NAME, (doc: Document) => {
+		const meshes = doc.getRoot().listMeshes();
+		for (const mesh of meshes) {
+			const srcPrims = mesh.listPrimitives();
+			for (const prim of srcPrims) {
```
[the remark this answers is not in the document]
donmccurdy: The simplest way to check and skip non-triangle primitives would be:

```javascript
if (prim.getMode() === Primitive.Mode.TRIANGLES) {
  ...
}
```

packages/functions/src/primitive-outline.ts
```diff
@@ -0,0 +1,135 @@
...
+		.createAccessor()
+		.setArray(new Float32Array(vertices))
+		.setType('VEC3')
+		.setBuffer(positionAccessor.getBuffer());
+
+	const edgePrim = document
+		.createPrimitive()
+		.setName(prim.getName() + '_edges')
+		.setMode(Primitive.Mode.LINES)
+		.setAttribute('POSITION', accesor);
```
donmccurdy: Perhaps ideally, all vertex attributes from the source primitive would be preserved in the output, but I recognize that's a fairly tricky change. For example, so that edges of a skinned mesh can be animated along with the rest of the mesh. Feel free to skip it and we can include it as a later improvement!

packages/functions/src/primitive-outline.ts
```diff
@@ -0,0 +1,135 @@
+import { type Document, Primitive, Triangle, type vec3 } from '@gltf-transform/core';
+import { vec3 as glvec3 } from 'gl-matrix';
+import { assignDefaults, createTransform } from './utils.js';
+
+const NAME = 'primitive-outline';
+
+export interface PrimitiveOutlineOptions {
+	thresholdAngel: number;
```
donmccurdy: Typo: `thresholdAngle`

packages/functions/src/primitive-outline.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+export const PRIMITIVE_OUTLINE_DEFAULTS: Required<PrimitiveOutlineOptions> = {
+	thresholdAngel: 0.05,
+};
+
+export function createPrimitiveOutline(_option: PrimitiveOutlineOptions = PRIMITIVE_OUTLINE_DEFAULTS): Primitive {
```
donmccurdy: Currently the convention in the repository is for transforms (functions passed to `document.transform(...)`) to have very concise names, and I'm undecided about how to expose this as a transform, I think it would require some ways to target particular meshes.

I'd prefer, at least for now, to remove the createPrimitiveOutline transform from this PR, and instead to export the `createEdgePrimitive` function as a standalone function taking a Primitive as input, like:

```typescript
export function createEdgePrimitive(prim: Primitive, thresholdRadians: number): Primitive {
  const graph = prim.getGraph();
  const document = Document.fromGraph(graph)!;
  ...
}
```

... does that sound OK?

packages/functions/src/primitive-outline.ts
```diff
@@ -0,0 +1,135 @@
...
+	const indices = prim.getIndices()?.getArray();
+	const indexCount = indices ? indices.length : positions!.length / 3;
+	const precisionCount = 4;
+	const precision = Math.pow(10, precisionCount);
+
+	const thresholdAngle = options.thresholdAngle;
+
+	const indexArr = [0, 0, 0];
+	const vertKeys = ['a', 'b', 'c'];
+	const triangle = new Triangle();
```
donmccurdy: nit: unused variable

donmccurdy (commented):
Thank you @WallanceLee, and sorry for the slow review! I'd be excited to get this feature merged, and have left some feedback on the PR in comments below. If any of the suggestions are confusing or feel like more than you can do right now, let me know and we'll figure something out. :)

WallanceLee:
OK, I will try my best to optimize the code.

packages/functions/src/primitive-outline.ts
```diff
@@ -0,0 +1,135 @@
...
+
+	const accesor = document
+		.createAccessor()
+		.setArray(new Float32Array(vertices))
+		.setType('VEC3')
+		.setBuffer(positionAccessor.getBuffer());
+
+	const edgePrim = document
+		.createPrimitive()
+		.setName(prim.getName() + '_edges')
```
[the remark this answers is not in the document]
WallanceLee: > Surprisingly, glTF primitives do not have names, so this line has no effect on export and can be removed. The `.setName()` method in glTF Transform exists only as inherited from the Property class.

Do you need to solve this btw?

PR #1738, opened 2025-10-22
Feat/cdsv 995

SIlin-CDS:


donmccurdy:
Hi @SIlin-CDS, thanks for opening the PR! Could you provide some background about the changes you're making here, and what issues you're trying to solve? A couple concerns:

- I don't think we can change the behavior of `getBounds` to accommodate Draco compression. If Draco compression requires quantization improvements though, we could discuss what those changes should be.
- I'd prefer not to inject custom `batch.extras` with the instance() function, since everyones' needs may be a bit different. I would recommend creating your own custom version of this instance() function instead, similar to my response in https://github.com/donmccurdy/glTF-Transform/pull/1719#issuecomment-3404312237.
- Similarly I would suggest just customizing a copy of `partition()` for your needs. It's a very simple function and should be easier to customize as needed, rather than to add complexity options for different use cases.

For these or other changes, please feel very welcome to open issues discussing problems you're trying to solve, and we can discuss options there.

Issue #1722, opened 2025-09-27
dedup: animation samplers


kzhsw:
**Is your feature request related to a problem? Please describe.**
Currently [dedup](https://github.com/donmccurdy/glTF-Transform/blob/main/packages/functions/src/dedup.ts) is only available for accessors, meshes, images, materials, and skins, but in gltf [spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#animations), samplers are defined to be able to be reused, this should help in case of models with duplicates in samplers.

**Describe the solution you'd like**
Add a `PropertyType.ANIMATION_SAMPLER` to `options.propertyTypes` of `dedup`, that enables animation sampler deduplication inside every gltf animation.

**Describe alternatives you've considered**

**Additional context**
<details><summary>Attach some code here</summary>
<p>

```typescript

function hashAnimationSampler(sampler: AnimationSampler): string {
	const hashKeys: (string | number | boolean)[] = [sampler.getInterpolation()];
	const input = sampler.getInput();
	if (input) {
		hashKeys.push(
			input.getCount(),
			input.getType(),
			input.getComponentType(),
			input.getNormalized(),
			input.getSparse(),
		);
	}
	const output = sampler.getOutput();
	if (output) {
		hashKeys.push(
			output.getCount(),
			output.getType(),
			output.getComponentType(),
			output.getNormalized(),
			output.getSparse(),
		);
	}
	return hashKeys.join(':');
}

function dedupAnimationSamplers(document: Document): void {
	const logger = document.getLogger();
	const root = document.getRoot();
	const animations = root.listAnimations();
	const samplerMap = new Map<string, AnimationSampler[]>();
	const duplicates = new Map<AnimationSampler, AnimationSampler>();
	let duplicateCount = 0, totalCount = 0;
	for (let i = 0, length = animations.length; i < length; i++) {
		const animation = animations[i];
		const samplers = animation.listSamplers();
		totalCount += samplers.length;
		for (let j = 0, samplersLength = samplers.length; j < samplersLength; j++) {
			const sampler = samplers[j];
			if (duplicates.has(sampler)) continue;
			const hashKey = hashAnimationSampler(sampler);
			const values = samplerMap.get(hashKey);
			if (!values) {
				samplerMap.set(hashKey, [sampler]);
				continue;
			}
			let isDuplicate = false;
			for (let k = 0, valueCount = values.length; k < valueCount; k++) {
				const value = values[k];
[...]

Issue #1708, opened 2025-09-08
dedup: Group material before check


kzhsw:
**Is your feature request related to a problem? Please describe.**
When running `dedup` on [NodePerformanceTest.glb](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/NodePerformanceTest/glTF-Binary/NodePerformanceTest.glb), it took 179s on `dedupMaterials`

<img width="1913" height="912" alt="Image" src="https://github.com/user-attachments/assets/efc7018a-65ed-4976-b565-f785a816ca52" />

<img width="832" height="864" alt="Image" src="https://github.com/user-attachments/assets/937abe90-bb4a-4ccb-a1c6-ac0bcc5393e8" />

**Describe the solution you'd like**
As comment states `dedupMaterials` runs a `O(n²)` compare, so it might help to group materials before compare. Pick some props that can diff from materials (mode, colors, textures, extensions), and hash them as the group key, then run `O(n²)` compare within each group.

**Describe alternatives you've considered**
Add an option to `dedup` that skips `dedupMaterials` if there are too many materials.

**Additional context**

kzhsw:
> what use case do you have in mind for optimizing glTF files containing ~10,000 materials?

Mostly models converted from other formats, FBX, STEP, IFC, and others.

> Computing a hash on all the material's properties, and its extensions' properties, would probably be more ideal and more general, but does require some additional work.

Yeah, hashing all props would clearly result in hashes with better quality, but might make hashing slower. Also, extensions should be considered.

Issue #1527, opened 2024-10-13
Memory overhead in scenes with many objects


donmccurdy:
_tl;dr — Scenes with many (1M+) meshes/nodes/accessors cause higher than expected memory cost and potential OOM_

While investigating a [recent question](https://blender.stackexchange.com/q/325519/43930) on the Blender Stack Exchange, I noticed that a glTF file linked from the question[^1] caused an OOM when running `gltf-transform inspect`. The file contained 200 MB of uncompressed binary data, which isn't a problem, but more importantly some 500,000 distinct mesh primitives. If the Node.js memory limit is around 4GB, that implies a memory cost of something like 8kb per mesh primitive. I can't easily measure memory allocation causes in Node.js developer tooling (that I'm aware of?), but the number looked higher than expected.

I ran further tests by printing heap size incrementally during loading. Heap size went from nominal to 3.7GB after initializing accessors, and hit OOM while processing mesh primitives. Note, this is not the binary vertex data, which has very predictable binary size, but the property graph organizing the scene/nodes/meshes/accessors on top of that data. On a hunch, I next tried to disable all event dispatchers and event listeners (managed by [property-graph](https://github.com/donmccurdy/property-graph/blob/main/src/event-dispatcher.ts)). This reduced heap size after initializing accessors to 2.4GB, and the scene loaded. Without events, some graph edges are never connected to the [Root](https://gltf-transform.dev/modules/core/classes/Root), so this leaves us with two plausible causes:

1. The events/listeners (and associated closures) cost more memory than expected
2. The memory cost is due to graph size, not events/listeners, and disabling the events reduced memory because another 1.5–2M graph edges were never connected

My feeling is that (1) is easier to fix, but (2) is more likely, and some more investigation will be needed to determine if one of these, or something else, is the cause.

[^1]: cad_machine.gltf, 7M vertices, 0.5M mesh primitives, 1.3M accessors

donmccurdy:
Results from memory profiling in Chrome devtools:

1. **Objects, 20%:** 8.2M objects, all reported appear to be _primitive_ objects. Mostly, just seeing the `$attributes` for Property instances here. A typical accessor holds about 120 bytes in the `$attributes` entry. I see a non-trivial number of _empty_ objects: `.extras={}`, `.extensions={}`, `.listeners={}` or `.listeners={dispose}`, etc. These appear to cost a flat 56 bytes, both `{}` and `{dispose}`. 
[...]

PR #1757, opened 2025-11-08
feat(functions): Add hashProperty(...) function

donmccurdy:
`hashProperty(prop)` returns a hash computed from all attributes and references held by this property, excluding the 'skip' set. Hash collisions are rare, but possible, so hashes alone should not be used to check for equality, but can be combined with [`Property#equals`](http://localhost:3000/modules/core/classes/Property#equals).

See also: https://en.wikipedia.org/wiki/Merkle_tree

Related:

- https://github.com/donmccurdy/glTF-Transform/issues/1722
- https://github.com/donmccurdy/glTF-Transform/issues/1708
- https://github.com/donmccurdy/glTF-Transform/issues/1527
- https://github.com/donmccurdy/glTF-Transform/issues/881

#### PR Dependency Tree

* **PR #1757** 👈
  * **PR #1760**
    * **PR #1761**

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)


kzhsw:
Would a "shallow hash" also be added here?
A "shallow hash" hashes references instead of values for things like `RefSet`, `RefMap`.
It can be implemented using a WeakMap, or add a counter when a `Property` constructed, can reduce a lot of compute when deep hash of values not needed.

<details><summary>WeakMap impl</summary>
<p>

```js
  // WeakMap is the key to being GC-friendly. It won't prevent garbage
  // collection of objects that are no longer referenced elsewhere.
  const cache = new WeakMap();
  // A simple counter to generate unique IDs for new objects.
  let nextId = 0;
  /**
   * Hashes a value.
   * @param {object} value The value to hash.
   * @returns {number} The hash value.
   */
  function shallowReferenceHash(value) {
    // Check if we have already hashed this exact object reference.
    if (cache.has(value)) {
      return cache.get(value);
    }
    // If it's a new object reference, generate a new ID, store it, and return it.
    const hash = nextId++;
    cache.set(value, hash);
    return hash;
  };
```


</p>
</details>

donmccurdy:
Might need that shallow hash after all! Here's a case where `toHash` is currently failing:

```javascript
test('toHash - circular reference', async (t) => {
	// Create a circular reference, nodeA -> nodeB -> skin -> nodeA.
	const meshNode = document.createNode('Mesh');
	const skeletonNode = document.createNode('Skeleton').addChild(meshNode);
	const skin = document.createSkin().addJoint(skeletonNode).setSkeleton(skeletonNode);
	meshNode.setMesh(document.createMesh()).setSkin(skin);

	t.is(typeof meshNode.toHash(), 'number', 'computes hash');
});

// -> Maximum call stack size exceeded
```

That's tricky, because circular references are rare but possible; the only example I know of is skinning. Ideas:

- (a) make toHash shallow only, don't follow property references
- (b) special-case the toHash calculation in the Skin class
- (c) ...figure out a [general-purpose solution](https://stackoverflow.com/questions/4540600/how-to-hash-and-check-for-equality-of-objects-with-circular-references) for hashing graph nodes circular references?

donmccurdy:
Resolved with a `depth` parameter. For purposes of `dedup()`, this requires some care when choosing how deeply to compare objects, but can be handled automatically.

Next steps:

- test that hash collision ratio is reasonable / small
- update benchmarks
- final tests and merge PR stack

PR #1770, opened 2025-11-28
fix (texture-basisu): edge case causes crashing on non textured model

chrisdaid:
Non-textured models that require basisu crashes with "Cannot read properties of undefined (reading 'forEach')"


donmccurdy:
Thanks, good catch @chrisdaid!

PR #1779, opened 2025-12-09
Filter or exclude by custom material properties in `toktx`

FrostKiwi:
This PR allows for to include or exclude textures from `toktx` in the `document.transform` step. It works specifically with blender custom properties (`extras` in gltf lingo) and assumes them to be booleans.

eg.:
```js
toktx({
    encoder: sharp,
    mode: Mode.ETC1S,
    slots: /baseColor|emissiveTexture|metallicRoughnessTexture/,
    excludeCustomTextureBool: "palletize", /* Exclude textures marked as palletized */
    compression: 3,
    filter: Filter.BOX,
}),
```

I copied the code from https://github.com/donmccurdy/glTF-Transform/blob/f937abbf27d3fa53398e3c2a2ccde5c00634b82a/packages/functions/src/join.ts#L152 to do the matching. I don't expect this PR to be merged as is, this is to kick-off a train of thought on how to include custom properties in the filtering. Thus also no tests yet in this PR.

## Motivation
Similar to the palletize function, I already pre-palletize in Blender and tag those materials for exclusion in texture compression and setting those to Nearest Neighbor, similar to function `pallete`. But gltf-transform `toktx` has no way to check for custom properties, so right now I have to do some gymnastics with checking custom property with the JS API, then adding a Texture name & URI prefix only then to have the `pattern` RegEx do the exclusion.

Right now I mark palletized materials:
```js
if (extras.palletized !== true) continue;
```
to then prefix them with this cursed setup:
```js
const PALLETIZED_PREFIX = "PALLETIZED_";
...
if (!currentURI.startsWith(PALLETIZED_PREFIX)) {
    /* Ensure URI has an extension so toktx can infer mime type */
    const uriWithExt = currentURI || "texture.png";
    texture.setURI(PALLETIZED_PREFIX + uriWithExt);
}
if (!currentName.startsWith(PALLETIZED_PREFIX))
    texture.setName(PALLETIZED_PREFIX + currentName);
```
only to then exclude them from the processing
```js
/* ETC1S compression for color textures (excluding palletized) */
toktx({
    encoder: sharp,
    mode: Mode.ETC1S,
    slots: /baseColor|emissiveTexture|metallicRoughnessTexture/,
    pattern: /^(?!PALLETIZED_)/, /* Exclude textures marked as palletized */
    compression: 3,
    filter: Filter.BOX,
}),
```

Either way, filtering by custom property seems like a natural extension to anything gltf processing related.
# Issues
- Technically, what is needed is to match by custom property of the texture, not the material. But even though [according to the Blender wiki](https://docs.blender.org/manual/en/latest/files/custom_properties.html) "All [data-blocks types](https://docs.blender.org/manual/en/latest/files/data_blocks.html#data-system-datablock-types) can have custom properties" and according to the [data-block type wiki page](https://docs.blender.org/manual/en/latest/files/data_blocks.html#data-system-datablock-types) `Image` is a `data-block type`, there exists no way in the Blender UI to set custom properties and setting it via the API doesn't work either, as the gltf exporter ignores it. Both things I'll try to look into
- `toktx` already has RegEx in the form of `pattern`, so maybe that feature could be connected to custom property things, but I didn't come up with a good way and resorted to `excludeCustomTextureBool` and `matchCustomTextureBool`
- I assume it to be a boolean, but really ideally we should be able to match strings and numbers for whatever other use-cases are doing

### DIFF
--- a/packages/cli/src/transforms/toktx.ts
+++ b/packages/cli/src/transforms/toktx.ts
@@ -4,6 +4,7 @@ import {
 	FileUtils,
 	type ILogger,
 	ImageUtils,
+	Material,
 	type Texture,
 	TextureChannel,
 	type Transform,
@@ -75,6 +76,10 @@ interface GlobalOptions {
 	 * Passing a string (glob) is deprecated; use a RegExp instead.
 	 */
 	slots?: RegExp | null;
+	/** Include only textures where a parent material has this custom property set to true. */
+	matchCustomMaterialBool?: string | null;
+	/** Exclude textures where a parent material has this custom property set to true. */
+	excludeCustomMaterialBool?: string | null;
 	/** Interpolation used for generating mipmaps. Default: 'lanczos4'. */
 	filter?: string;
 	filterScale?: number;
@@ -121,6 +126,8 @@ const GLOBAL_DEFAULTS: Omit<GlobalOptions, 'encoder' | 'mode'> = {
 	filterScale: 1,
 	pattern: null,
 	slots: null,
+	matchCustomMaterialBool: null,
+	excludeCustomMaterialBool: null,
 	// See: https://github.com/donmccurdy/glTF-Transform/pull/389#issuecomment-1089842185
 	jobs: 2 * NUM_CPUS,
 	cleanup: true,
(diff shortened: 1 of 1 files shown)


donmccurdy:
Closing for now, but happy to reopen if this is still something you're interested in continuing! :)

Issue #1448, opened 2024-07-03
KHR_animation_pointer extension not preserved


heeen:
**Describe the bug**
importing a file with KHR_animation_pointer and re-serializing it breaks said extension

**To Reproduce**
WebIO with ALL_EXTENSIONS registered
import document through io.readJSON
export document through io.writeJSON

**Expected behavior**
it is expected that the gltf-transform without explicit changes is idempotent. 

original file contains 
```
"extensionsUsed": [
    "KHR_animation_pointer"
  ],
```
and animations
```
  "animations": [
    {
      "channels": [
        {
          "sampler": 0,
          "target": {
            "path": "pointer",
            "extensions": {
              "KHR_animation_pointer": {
                "pointer": "/nodes/2/rotation"
              }
            }
          }
        }
      ],
      "samplers": [
        {
          "input": 13,
          "interpolation": "LINEAR",
          "output": 14
        }
      ],
      "name": "SpinPlatform"
    },
    {
      "channels": [
        {
          "sampler": 0,
          "target": {
            "path": "pointer",
            "extensions": {
              "KHR_animation_pointer": {
                "pointer": "/nodes/5/rotation"
              }
            }
          }
        },
        {
          "sampler": 1,
          "target": {
            "path": "pointer",
            "extensions": {
              "KHR_animation_pointer": {
                "pointer": "/nodes/3/translation"
              }
            }
          }
        }
      ],
```

resulting does not contain extensionsUsed. animations look like:
```
"animations": [
        {
            "name": "SpinPlatform",
            "samplers": [
                {
                    "input": 13,
                    "output": 14,
                    "interpolation": "LINEAR"
                }
            ],
            "channels": [
                {
                    "sampler": 0,
                    "target": {
                        "path": "pointer"
                    }
                }
            ]
        },
        {
            "name": "animate the inner box",
            "samplers": [
                {
                    "input": 15,
                    "output": 16,
                    "interpolation": "LINEAR"
                },
                {
                    "input": 17,
                    "output": 18,
                    "interpolation": "LINEAR"
                }
            ],
            "channels": [
                {
                    "sampler": 0,
                    "target": {
                        "path": "pointer"
                    }
                },
                {
                    "sampler": 1,
                    "target": {
                        "path": "pointer"
                    }
                }
            ]
        }
    ],
```

**Versions:**
 - Version: 4.0.2
 - Environment: Browser

heeen:
[...]

PR #1820, opened 2026-05-27
Feat/khr animation pointer

bbhieu1201:


issacclee:
interesting, may I enquiry what's the current status of this implementation?

donmccurdy:
@bbhieu1201 I'd be interested to hear your goals and status for this PR as well!

From my perspective, support for `KHR_animation_pointer` in glTF Transform is important, but complex, and has been beyond the (personal) development time I've had available to spend here recently.

I think this PR is a good start (thank you for including unit tests, in particular!) but leaves a lot unfinished in the scope of animatable properties, see:

- https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/ObjectModel.adoc

While I understand that initial support for the animated properties included in this PR may be sufficient for your own goals, by merging this I'd be in the position of maintaining it and responding to bug reports when other aspects of the extension do not work, which may require substantial development. 

Additionally many of the functions in `@gltf-transform/functions` may impact animated properties, and considering and testing those cases would require some investigation.

So, I think that either supporting the extension's features more comprehensively, or an agreed architectural plan for how this would be handled moving forward, are probably necessary before this PR can be merged. In the meantime, of course, it is also possible to maintain an extension separately from the glTF Transform repo and to register it with `io.registerExtensions(...)` as needed.

Related:

- #1448

donmccurdy:
Would be happy to reopen and review this or another PR, but support for KHR_animation_pointer will require discussing and/or addressing the questions about scope and testing above.

PR #1835, opened 2026-08-01
feat(cli): add --resize to etc1s and uastc

### LINKED ISSUE #1776: CLI 'etc1s' and 'uastc' commands should expose --resize flag
When running the 'optimize' CLI command with KTX2 compression enabled by the `--texture-compress` flag, texture resizing (if enabled by `--resize`) happens automatically within the toktx() calls. When running the `etc1s` or `uastc` commands separately, `--resize` isn't currently exposed, and so you'd need to call the `resize` function separately. 

My intention is that 'optimize' be an easy-to-use, simplified, and opinionated pipeline composed of parts from `@gltf-transform/functions`, and that it's possible — when those simplifications and opinions aren't a good fit for a particular user — to drop down into composing the more configurable and more advanced commands and functions underneath it.

That being the case, the `etc1s` and `uastc` commands should expose all features that they offer through the `optimize` command, and `--resize` is currently missing.

Guflly:
Adds `--resize <size>` to the standalone `etc1s` and `uastc` commands and passes square dimensions through to `toktx()`.

Includes CLI coverage for both commands. Tests: CLI tests and build.

Closes #1776.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1341,6 +1341,9 @@ normal maps and ETC1S for other textures, for example.`.trim(),
 		validator: Validator.STRING,
 	})
 	.option('--slots <slots>', 'Texture slots to include (glob)', { validator: Validator.STRING })
+	.option('--resize <size>', 'Maximum texture dimensions, in pixels.', {
+		validator: Validator.NUMBER,
+	})
 	.option('--mipmaps <bool>', 'Generate mipmaps.', {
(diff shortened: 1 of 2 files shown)


donmccurdy:
It is my impression that this is an automated PR from an automated account. For it to be worthwhile for me to spend time reviewing this the PR, I would need to have confidence that there's a human on the other side of the conversation, using this feature in glTF Transform.

I'm closing the PR accordingly. Apologies if I've misunderstood.

PR #1850, opened 2026-08-27
chore(build): Add type check (tsc) step to build

### LINKED ISSUE #1780: Include a 'tsc' check in CI
It's possible for TypeScript to locally detect errors (with typescript-language-server) that are not detected in the build (with Microbundle).  Most recently, in https://github.com/zeux/meshoptimizer/pull/997. It might be best to include a call to `tsc` in the CI process so that TypeScript does checks explicitly. I'm less sure about why it fails, or doesn't, within Microbundle.

The /docs package has some dependencies that are not compatible with `"moduleResolution": "nodenext"`, so I might need to change the moduleResolution setting for that package, or update those dependencies.

donmccurdy:
- Fixes #1780

#### PR Dependency Tree

* **PR #1850** 👈

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
       - run: corepack enable
       - run: yarn install --immutable
       - run: yarn build
+      - run: yarn build:tsc
 
       # Test.
       - run: yarn test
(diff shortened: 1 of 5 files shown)


donmccurdy:
Pending:

- https://github.com/toji/watlas/pull/3

PR #1852, opened 2026-08-30
fix(coverage): enable source maps so c8 reports on source, not bundled dist

### LINKED ISSUE #1210: Code coverage drops some sourcemaps
Running locally, code coverage reports on the bundled files for cli/ and core/, and for the source files for functions/ and extensions/. I'd rather it report on source files in all cases, but I'm not sure why the difference exists.

- https://app.codecov.io/gh/donmccurdy/glTF-Transform/tree/main/packages

Mazen-2010-afk:
Fixes #1210. Code coverage reports on bundled dist/ files instead of source .ts files.

Changes: tsdown.config.ts adds sourcemap:true; .c8rc adds all:true.

Closes #1210

### DIFF
--- a/.github/.c8rc
+++ b/.github/.c8rc
@@ -1,4 +1,5 @@
 {
+	"all": true,
 	"exclude": [
 		"packages/*/vendor",
 		"packages/*/test",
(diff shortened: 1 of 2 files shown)


donmccurdy:
Thanks @Mazen-2010-afk — testing this locally and it looks like building sourcemaps with tsdown is enough. Adding `"all": true` to c8rc brings in a lot of files (e.g. node_modules/ dependencies) that don't need to appear in the coverage report. Possibly best to omit that part of the change?

PR #1853, opened 2026-09-03
fix(view): enable flatShading on implicit default material

### LINKED ISSUE #1686: Model with no material and normals renders black
**Describe the bug**

I have this model of a character: https://cdn.shopify.com/3d/models/1d072fba58a00408/white_mesh_4_.glb

It doesn't have a material or a normals attribute.

When I read it and display it like this:

```
  const io = new WebIO()
    .registerExtensions(ALL_EXTENSIONS)
    .registerDependencies({
      "draco3d.encoder": await new DracoEncoderModule(),
      "draco3d.decoder": await new DracoDecoderModule(),
    });
  const originalDocument = await io.read(url);
  const modifiedDocument = cloneDocument(originalDocument);

  // Create a live view of the modified document
  const modifiedDocumentView = new DocumentView(modifiedDocument);
  const sceneDefinition = modifiedDocument.getRoot().getDefaultScene()!;
  const scene = modifiedDocumentView.view(sceneDefinition);

  // Render scene using React-Three-Fiber
  // My scene has an environment map
  // ... 
```

The model renders black:

Image

If I print its material, I can see it's this:

```
export const DEFAULT_MATERIAL: MeshStandardMaterial = new MeshStandardMaterial({
	name: '__DefaultMaterial',
	color: 0xffffff,
	roughness: 1.0,
	metalness: 1.0,
});
```

That's white, so the reason our model is black is because it doesn't have any normals.

Correct me if I'm wrong, but I thought that if a model has no normals, Three.js generates them using this function: https://threejs.org/docs/#api/en/core/BufferGeometry.computeVertexNormals

If I do this in my code:

```
  const { scene } = useThree();

  useEffect(() => {
    if (scene) {
      scene.traverse((child: any) => {
        if (child.isMesh && child.material.name === "__DefaultMaterial") {
          child.geometry.computeVertexNormals();
        }
      });
    }
  }, [scene]);
```

The model is rendered with proper lighting:

Image

The same is the case if I load the model with useGLTF:

```
import { useGLTF } from "@react-three/drei";
import * as THREE from "three";
import { GLTF } from "three-stdlib";

type GLTFResult = GLTF & {
  nodes: {
    tmpfp5to6beply: THREE.Mesh;
  };
  materials: {};
};

export function Doorman(props: JSX.IntrinsicElements["group"]) {
  const { nodes, materials } = useGLTF(
    "https://cdn.shopify.com/3d/models/1d072fba58a00408/white_mesh_4_.glb"
  ) as GLTFResult;

  console.log(nodes.tmpfp5to6beply.geometry.attributes);

  return (
    <group {...props} dispose={null}>
      <mesh
        geometry={nodes.tmpfp5to6beply.geometry}
        material={nodes.tmpfp5to6beply.material}
      />
    </group>
  );
}

useGLTF.preload(
  "https://cdn.shopify.com/3d/models/1d072fba58a00408/white_mesh_4_.glb"
);
```

So I guess my question is: should the live-view of the scene provided by glTF-Transform compute vertex normals?

As an aside, I wanted to mention that the model I linked here causes both https://gltf-viewer.donmccurdy.com/ and https://gltf.report/ to crash.

Thank you for any information.

**Versions:**
 - Version: glTF-Transform 4.2.0
 - Environment: Chrome, Vite, R3F

simonyang08:
Fixes #1686

When a primitive has no source material, `PrimitiveSubject` substitutes the shared `DEFAULT_MATERIAL` directly and never routes through `MaterialPool`, so the adaptations that MaterialPool normally applies for primitives without `NORMAL` are silently skipped. A mesh with an unmodified `MeshStandardMaterial` and no normal data renders black.

### Change

Mirror the MaterialPool flatShading adaptation for primitives that fall back to the implicit default material:

- `flatShading = !hasNormals` on `DEFAULT_MATERIAL` (matching `MaterialPool.ts`'s `flatShading = params.useFlatShading`), following the flat-normals requirement from the glTF spec noted in the issue thread;
- re-applied whenever the attribute observer fires, so adding/removing `NORMAL` on a primitive still using the default material keeps the shading mode correct;
- because `flatShading` is a shader-compilation parameter and the default material is a shared singleton, `needsUpdate` is set on every flip — otherwise already-compiled programs keep the stale shading mode.

### Known limitation

Two primitives in the same model that both use the implicit default material — one with `NORMAL`, one without — share a single material instance, so the flatShading flag reflects whichever primitive was adapted last (same behavior as THREE.GLTFLoader's shared-material path, but worth stating explicitly). A complete fix would route default-material primitives through MaterialPool's variant system; happy to attempt that if preferred.

`InstancedMeshSubject` also binds `DEFAULT_MATERIAL` unconditionally and likely needs the same adaptation; left out of this focused fix, can follow up.

### Test

New regression test in `packages/view/test/PrimitiveSubject.test.ts`: a primitive with only `POSITION` gets `flatShading=true` on the default material; toggling `NORMAL` off/on flips the flag accordingly. The shape assertions fail on `main`.

`node --test`: 412 tests on this branch, same 12 pre-existing Windows path-resolution failures as `main` (unrelated fixture-path issue), plus this new passing test. biome clean on changed files.

### DIFF
--- a/packages/view/src/subjects/PrimitiveSubject.ts
+++ b/packages/view/src/subjects/PrimitiveSubject.ts
@@ -11,6 +11,7 @@ import {
 	LineSegments,
 	type Material,
 	Mesh,
+	MeshStandardMaterial,
 	Points,
 	SkinnedMesh,
 } from 'three';
@@ -41,6 +42,13 @@ export class PrimitiveSubject extends Subject<PrimitiveDef, MeshLike> {
 		this.material.subscribe((material) => {
 			if (this.value.material !== material) {
 				this.value.material = material || DEFAULT_MATERIAL;
+				// When the primitive has no source material, apply the same
+				// flatShading adaptation that MaterialPool would normally
+				// perform, so the implicit default material doesn't render black.
+				// See https://github.com/donmccurdy/glTF-Transform/issues/1686
+				if (!material) {
+					PrimitiveSubject.applyDefaultMaterialAdaptations(this.def, this.value);
+				}
 				this.publishAll();
 			}
 		});
@@ -57,6 +65,12 @@ export class PrimitiveSubject extends Subject<PrimitiveDef, MeshLike> {
 			}
 			for (const key in nextAttributes) {
 				geometry.setAttribute(semanticToAttributeName(key), nextAttributes[key]);
+			}
+			// If the primitive uses the implicit default material and the
+			// NORMAL attribute was removed/changed, re-apply adaptations.
+			// See https://github.com/donmccurdy/glTF-Transform/issues/1686
+			if (this.value.material === DEFAULT_MATERIAL) {
+				PrimitiveSubject.applyDefaultMaterialAdaptations(this.def, this.value);
 			}
 			this.publishAll();
 		});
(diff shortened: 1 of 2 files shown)


simonyang08:
Hi @donmccurdy — I noticed the PR was closed without feedback. Was the flatShading approach itself the concern, or was it something else (e.g. the `computeVertexNormals` mention in the description, which I had already dropped from the implementation)? Happy to rework it, or to close #1686 as wontfix if you'd rather not take this direction. Either way, thanks for maintaining the library.

donmccurdy:
The PR is closed because it appears to be part of an automated, bulk batch of AI-generated PRs. While I'm happy to work with a human to refine and merge a PR, coaching an LLM toward a better implementation is not a good use of my time.

In general: comments, issues, and PR descriptions written by AI may be closed at my discretion. Communication on this repository should be between humans.

